PR #20, opened 2017-06-26
RFC: Adds function to replace html contents and run scripts

JobJob:
Adds a function that takes an string of html to replace the contents of an element,
and ensures any scripts in the html string get run. This can be done using `element.innerHTML = htmlstr` but any scripts won't get run.

Used in InteractNext to auto-update display of Observables here: https://github.com/JuliaGizmos/InteractNext/blob/master/src/observablesIO.jl#L32

### DIFF
--- a/assets/basics/webio.js
+++ b/assets/basics/webio.js
@@ -150,6 +150,32 @@ function sendNotSetUp()
     console.error("WebIO.sendCallback not set up")
 }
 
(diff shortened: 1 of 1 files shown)


shashi:
Looks good! This is a Utility function which doesn't seem to have anything to do with WebIO itself. Once we have enough of these, maybe we should move them out to a Utilities module.

Cheers!

PR #21, opened 2017-07-07
Widget lifecycle notifications, register_renderable, and attrUtils

JobJob:
See the commit messages for details

### DIFF
--- a/assets/basics/node.js
+++ b/assets/basics/node.js
@@ -281,6 +281,7 @@ function createWidget(ctx, data) {
 
     depsPromise.then(function (deps) {
         appendChildren(subctx, fragment, data.children);
+        WebIO.send(subctx, "widget_created", {})
     })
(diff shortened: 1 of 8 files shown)


assets/basics/node.js
```diff
@@ -260,6 +262,8 @@ function createWidget(ctx, data) {
     var subctx = WebIO.makeWidget(data.instanceArgs.id, ctx.data,
                              ctx.sendCallback, fragment, command_funcs, observables);
 
+    notifyall("preDependencies", {}, subctx, command_funcs)
```
shashi: Why do we need `notifyall` ? Can't we use `WebIO.setval` or something?

JobJob:
Closing in favour of #22 - that one adds nested scripts and removes the `notifyall` which we'll implement differently.

PR #27, opened 2017-08-19
Fix JSONification of interpolated `@js` function code 

JobJob:
(This PR is only for the last commit, merge #25 then #26 first)

Previously, interpolated functions were generally in double quotes so they weren't `eval`ed correctly. This meant you couldn't define say Dicts with callback functions as values to include into later code.

Example:
```
using WebIO
jsobj = Dict(:hi=>@js function() hi end)
jscode = @js function()
    $jsobj
end
jscode.s |> println
```
Before: `(function (){return {"hi":"(function (){ alert("hi") })"}})`
After:    `(function (){return {"hi":(function (){ alert("hi") })}})`

Note the double quotes around the nested "hi" function.

edit: Also has a small fix for named functions without arguments

### DIFF
--- a/src/syntax.jl
+++ b/src/syntax.jl
@@ -43,15 +43,36 @@ end
 using JSON, MacroTools
 export JSString
 
-jsexpr(io, x) = JSON.print(io, x)
-
 immutable JSString
     s::String
 end
 
 Base.:(==)(x::JSString, y::JSString) = x.s==y.s
 
+const JSONContext = JSON.Writer.StructuralContext
+const JSONSerialization = JSON.Serializations.CommonSerialization
+
+struct JSEvalSerialization <: JSONSerialization end
+
 JSON.lower(x::JSString) = x.s
+
+const verbose_json = Ref(false)
+
+# adapted (very slightly) from JSON.jl test/serializer.jl
+function JSON.show_json(io::JSONContext, ::JSEvalSerialization, x::JSString)
+    if verbose_json[]
+        first = true
+        for line in split(x.s, '\n')
+            !first && JSON.indent(io)
+            first = false
+            Base.print(io, line)
+        end
+    else
+        Base.print(io, x.s)
+    end
+end
+
+jsexpr(io, x) = JSON.show_json(io, JSEvalSerialization(), x)
 
 jsexpr(x) = JSString(sprint(jsexpr, x))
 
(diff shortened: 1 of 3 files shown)


src/syntax.jl
```diff
@@ -43,16 +43,37 @@ end
...
+function JSON.show_json(io::JSONContext, ::JSEvalSerialization, x::JSString)
+    if verbose_json[]
+        first = true
+        for line in split(x.s, '\n')
+            !first && JSON.indent(io)
+            first = false
+            Base.print(io, line)
+        end
+    else
+        Base.print(io, x.s)
```
shashi: This could be invalid JSON right?

Since in the next commit you use `jsexpr` to print a node, couldn't we just not do without overloading show_json?

shashi:
Neat, this is pretty useful!

src/syntax.jl
JobJob: Yes, it can be invalid json, because functions aren't part of the spec - it's valid js though.

It only overloads it when we use the `JSEvalSerialization()`

shashi: Ah, cool.

Okay I'll merge this then.

JobJob: The guys over at JSON.jl basically enable this show_json to take a different serialiser - see the bottom of here for our discussion: https://github.com/JuliaIO/JSON.jl/issues/170

and here for details about show json: https://github.com/JuliaIO/JSON.jl/pull/162#issuecomment-246133530

basically the tldr is this won't affect `JSON.json(::Node)`, will only affect when you explicitly use this JSEvalSerialization serialiser which is made in this PR somewhere

shashi:
So will this  https://github.com/JuliaGizmos/WebIO.jl/blob/master/assets/basics/node.js#L110 continue to work? Can that code be removed now that we just output code instead of strings for event handlers?

JobJob:
Yeah, I think it can be removed, but been rushing doing other stuff, so haven't looked into it.
Also, I should've written some tests 😛

JobJob:
After this (I think) you can easily add methods etc. to the Vue constructor with Vue.jl. Not sure if it was working before, been making lots of changes, can't remember what fixed what anymore 😄

shashi:
Alright! Loving the momentum. Vue was working before because Vue.js converts strings to code...

PR #29, opened 2017-08-24
Multi widget display and event and command handlers updated

JobJob:
Enables registering a callback in `show(n::Node, ...)`, which is a way for a single Widget to be displayed multiple times (there will be multiple js contexts associated with the same Julia Widget).

Also
* `after` now called with `apply` not `call`, meaning, for example, each dependency for ondependencies, is now passed as a separate argument, rather than the first argument holding an array.
* command and event handlers updated for new JSEvalSerializer JSON regime
* offjs method added (untested/unused in the end)
* Syntax for `+=` added for js macros

### DIFF
--- a/src/widget.jl
+++ b/src/widget.jl
@@ -150,7 +150,7 @@ function after(ctx::Widget, promise_name, expr)
     @evaljs ctx begin
         @var widget = this;
         this.promises[$promise_name].
-            then(val -> $expr.call(widget, val))
+            then(val -> $expr.apply(widget, val))
     end
 end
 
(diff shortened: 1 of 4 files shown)


src/widget.jl
```diff
@@ -150,7 +150,7 @@ function after(ctx::Widget, promise_name, expr)
     @evaljs ctx begin
         @var widget = this;
         this.promises[$promise_name].
-            then(val -> $expr.call(widget, val))
+            then(val -> $expr.apply(widget, val))
```
shashi: This is not very normal, what if I intended a promise to return an array?

I've added ondependencies function for the purpose of getting dependencies as separate arguments. Use that instead?

JobJob: Good point, reverted.

PR #32, opened 2017-09-10
Default rendering for all types and Observable Array Fixes

JobJob:
Adds default rendering for all types using richest show. This won't generally be used unless register_renderable(::Type) is called but can come in handy. E.g. InteractNext uses it to display auto-updating Observables

Fixes array valued observable updating. Turns out that in js `["a"] !== ["a"]` - this deals with that in the context of propagating observable updates

### DIFF
--- a/src/render.jl
+++ b/src/render.jl
@@ -1,0 +1,80 @@
+"""
+    render(x::MyType)
+
+Generic function that defines how a Julia object is rendered. Should return a
+`Node` object.
+"""
+function render end
+render(x::Node) = x
+render(x::Text) = dom"span"(x.content)
+render(x::String) = render(Text(x))
+render(::Void) = ""
+render(x::Any) =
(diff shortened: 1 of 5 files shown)


src/render.jl
```diff
@@ -0,0 +1,80 @@
...
+
+function richest_html(val)
+    topmime = richest_mime(val)
+    str_repr = stringmime(topmime, val)
+    res = (if topmime == MIME("text/html")
+        str_repr |> WebIO.encode_scripts
+    elseif topmime in map(MIME, ["image/svg+xml", "image/png", "image/jpeg"])
+        "<img src='data:image/png;base64,$str_repr'></img>"
+    elseif topmime == MIME("text/plain")
+        "<span>$str_repr</span>"
```
shashi: `pre` probably is better

JobJob: 👍

PR #44, opened 2018-02-12
WIP: API fixes

shashi:
- Widgets are now renderable on their own, this means you can write functions which return a widget, and use the same result to set some observable in the widget.

Demo  1:
```julia
function katex(txt)
    w = Widget(imports=[
            "https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-beta1/katex.min.js",
            "https://cdnjs.cloudflare.com/ajax/libs/KaTeX/0.9.0-beta1/katex.min.css"])

    w["text"][] = txt

    onimport(w, @js function (k)
        this.k = k
        this.container = this.dom.querySelector("#container")
        k.render($txt, this.container)
    end)

    onjs(w["text"], @js (txt) -> this.k.render(txt, this.container))

    w.dom = dom"div#container"() # widget is now mutable, so you can simply set the contents

    w
end
```
screenshot from 2018-02-12 17-30-45
- you can directly update the contents using `widget["text"][] = "new text"` (no need to maintain a separate reference to the observable)
screenshot from 2018-02-12 17-30-57

- Dependencies which need to be loaded synchronously can be put in a `Sync(["a.js", "b.js"])` wrapper. For examples `dependencies=[Sync(["a.js", "b.js"]), "c.js"]` will load `b.js` after `a.js` is loaded, but `c.js` in parallel with `a.js` and/or `b.js`.

Demo 2

```julia
using WebIO

function draggable_moment()
    w = Widget(imports=[Sync(
                ["https://code.jquery.com/jquery-1.12.4.js",
                 "https://code.jquery.com/ui/1.12.1/jquery-ui.js"]),
                "https://momentjs.com/downloads/moment.min.js"])
    onimport(w, @js function (jq, ui, moment)
      jq( "#draggable" ).draggable().text(moment("20180101", "YYYYMMDD").fromNow())
    end)

    w.dom = dom"div"(
        dom"div#draggable"("Drag me around", style=Dict("width" => "100px",
                :backgroundColor=>"orange", "height" => "100px", :padding=>"10px")),
        style=Dict("width" => "400px", "height" => "400px")
    )

    w
end

draggable_moment()
```
This will return a draggable box which contains the words "one month ago" created by moment.js. Here `jquery-ui.js` is loaded after `jquery.js`. but `moment.js` is loaded asynchronous to anything else. The `ondependencies` function gets 3 arguments (the respective modules), and is called when all the js files are loaded.

# Upcoming changes:
- [x] rename dependencies to `imports`. So `ondependencies` becomes `onimport`.
- [x] allow a vector of pairs of [name => jsurl] in dependencies so that `this.name` is set to the module automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,12 +234,12 @@ Note that `@js` just translates a Julia expression to the equivalent JavaScript,
 ## Loading JavaScript dependencies
 
 
-You can load dependencies by creating a Widget object and passing in `dependencies` argument.
+You can load dependencies by creating a Widget object and passing in `imports` argument.
 
 ```julia
-w = Widget(dependencies=["//cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js"])
+w = Widget(imports=["//cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js"])
 
-ondependencies(w, @js function (p5)
+onimport(w, @js function (p5)
     function sketch(s)
         s.setup = () -> s.createCanvas(640, 480)
 
--- a/assets/basics/node.js
+++ b/assets/basics/node.js
@@ -248,6 +248,9 @@ function doImports(widget, imp) {
                 return mods2
             })
         case "js":
+            var cfg = {paths: {}}
+            cfg.paths[imp.name] = imp.url
+            SystemJS.config(cfg)
             return SystemJS.import(imp.url).then(function (mod) {
                 if (imp.name) {
                     widget[imp.name] = mod
--- a/assets/basics/node.js
+++ b/assets/basics/node.js
@@ -192,58 +192,96 @@ function createDOM(ctx, data, parentNode) {
 
 require('../node_modules/systemjs/dist/system.js')
 
-function asyncloadJS(js) {
-    return SystemJS.import(js); // a promise
-}
+function doImports(imp) {
+    // this function must return a promise which resolves to a
+    // vector of modules or null values (null values for non-modules)
+    switch (imp.type) {
+        case "sync_block":
+            // all imports in this block must be loaded one after the other
+            // We create a single promise which accumulates the resolved
+            // values of all the input promises into a vector
+            var sync_imports = imp.data
 
-function doImports(imports) {
-    var promises = [];
-    for (var i=0, l=imports.length; i<l; i++) {
-        var imp = imports[i];
-        if (imp.type === undefined) {
-            console.warn("Can't load dependency ", imp)
-        } else {
-            switch (imp.type) {
-                case "js":
-                    promises.push(SystemJS.import(imp.url));
-                    break;
-                case "css":
-                    var cssId = imp.url.split( "/" ).join("-");
-                    if (!document.getElementById(cssId))
-                    {
-                        var head  = document.getElementsByTagName('head')[0];
-                        var link  = document.createElement('link');
-                        link.id   = cssId;
-                        link.rel  = 'stylesheet';
-                        link.type = 'text/css';
-                        link.href = imp.url
-                        link.media = 'all';
-                        head.appendChild(link);
-                    }
-                    break;
-                case "html":
-                    var p = new Promise(function (accept, reject) {
-                        var link = document.createElement('link');
-                        if ('import' in link) {
-                            link.rel = 'import';
-                            link.href = imp.url;
-                            link.setAttribute('async', '');
-                            link.onload = accept;
-                            link.onerror = reject;
-                            document.head.appendChild(link);
-                        } else {
-                            console.error("This browser doesn't support HTML imports. Cannot import " + imp.url);
-                            reject(imp);
-                        }
-                    });
-                    promises.push(p);
-                    break;
-                default:
-                    console.warn("Don't know how to load import of type " + imp.type);
+            if (sync_imports.length == 0) {
+                return undefined;
             }
-        }
+
+            // we chain promises to this empty promise
+            var p = new Promise(function (acc, rej) { acc([]); })
+
+            function makePromisClosure(sync_import) {
+                return function (vec) {
+                    // vec contains previously resolved modules
+                    var y = doImports(sync_import); // a promise
+                    var flatten = (sync_import.type == "sync_block" || sync_import.type == "async_block")
+                    return new Promise(function (acc, rej) {
+                        y.then(function (mod) {
+                            if (typeof(mod) === "undefined") {
+                            } else if (flatten) {
+                                acc(vec.concat(mod))
+                            } else {
+                                vec.push(mod)
+                                acc(vec)
+                            }
+                        })
+                    })
+                }
+            }
+            for (var i=0, l = sync_imports.length; i < l; i++) {
+                p = p.then(makePromisClosure(sync_imports[i]))
+            }
+            return p;
+        case "async_block":
+            var ps = imp.data.map(doImports)
+            return Promise.all(ps).then(function (mods) {
+                var mods2 = []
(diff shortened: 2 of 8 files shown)


assets/basics/node.js
```diff
@@ -192,58 +192,103 @@ function createDOM(ctx, data, parentNode) {
...
-                        link.id   = cssId;
-                        link.rel  = 'stylesheet';
-                        link.type = 'text/css';
-                        link.href = imp.url
-                        link.media = 'all';
-                        head.appendChild(link);
+            // we chain promises to this empty promise
+            var p = new Promise(function (acc, rej) { acc([]); })
+
+            function makePromisClosure(sync_import) {
```
JobJob: I Promis you made a typo

README.md
```diff
@@ -234,12 +234,12 @@ Note that `@js` just translates a Julia expression to the equivalent JavaScript,
 ## Loading JavaScript dependencies
 
 
-You can load dependencies by creating a Widget object and passing in `dependencies` argument.
+You can load dependencies by creating a Widget object and passing in `imports` argument.
```
JobJob: Add a note somewhere about A/Sync?

assets/basics/node.js
```diff
@@ -192,58 +192,103 @@ function createDOM(ctx, data, parentNode) {
...
+                    // vec contains previously resolved modules
+                    var y = doImports(widget, sync_import) // a promise
+                    var flatten = (sync_import.type == "sync_block" ||
+                                   sync_import.type == "async_block")
+
+                    return new Promise(function (acc, rej) {
+                        y.then(function (mod) {
+                            if (typeof(mod) === "undefined") {
+                            } else if (flatten) {
+                                acc(vec.concat(mod))
```
JobJob: Should this call `makePromiseClojure` for nested sync_imports? or do nested sync imports not make sense? Anyway could add that later.

assets/basics/node.js
```diff
@@ -192,58 +192,103 @@ function createDOM(ctx, data, parentNode) {
...
-    return SystemJS.import(js); // a promise
-}
+function doImports(widget, imp) {
+    // this function must return a promise which resolves to a
+    // vector of modules or null values (null values for non-modules)
+    switch (imp.type) {
+        case "sync_block":
+            // all imports in this block must be loaded one after the other
+            // We create a single promise which accumulates the resolved
+            // values of all the input promises into a vector
```
JobJob: ace 👍

src/widget.jl
```diff
@@ -21,91 +23,98 @@ An object which can send and receive messages.
...
 end
 
-function (::Type{Observable{T}}){T}(ctx::Widget, cmd, value; sync=nothing)
-    setobservable!(ctx, cmd, Observable{T}(value), sync=sync)
+function Base.getindex(w::Widget, key)
+    key = string(key)
+    if haskey(w.observs, key)
+        w.observs[key][1]
+    else
+        Observable{Any}(w, key, nothing)
```
JobJob: tricky!

assets/basics/node.js
shashi: They definitely make sense, but I would think they will have their own closures since they recursively call doImports right?

src/widget.jl
shashi: You mean not good?

This lets you write one less line per widget -- the one where you create the observable, but it also causes you to get stuck if you made a typo because you won't get an error saying that observable doesn't exist...

shashi:
I'm trying to port @aditbiswas1's React.js integration [gist](https://gist.github.com/aditbiswas1/5c2c93280e9081bdaf3b15c0cbef3600) to this setup using

```julia
deps = Sync(["https://cdnjs.cloudflare.com/ajax/libs/react/16.0.0/umd/react.development.js",
             "https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.0.0/umd/react-dom.development.js",
             "https://cdn.jsdelivr.net/npm/create-react-class@15.6.2/create-react-class.js"])

const HelloComponent = Dict(
    "render"=> WebIO.@js function()
        React.createElement("div", null, "Hello")
    end
)

const ondeps_fn = WebIO.@js function(React, ReactDOM, createReactClass)
    console.log(React, ReactDOM, createReactClass);
    @var helloComponent = createReactClass($HelloComponent);
    ReactDOM.render(
        React.createElement(helloComponent, null, null),
        this.dom,
    )
end


widget = Widget(imports=deps)
onimport(widget, ondeps_fn)
widget.dom = dom"div#root"()

widget
```

But I get this super weird error:

screenshot from 2018-02-13 11-21-21

react-dom.development.js seems to be trying to load http://localhost:8888/notebooks/react  (or `react` relative path), sure enough that URL returns an "not found" HTML page, which causes the syntax error. I don't know what's wrong, and it worries me. Anyone has any guesses?

aditbiswas1:
i'll take a look at this again, are you getting the component to render ?
i might have seen this issue at some point trying to make that gist work

shashi:
No, it fails before it can run `onimport` handler

shashi:
@aditbiswas1 I encountered this when trying to load jQuery and jQueryUI (as in my example in the OP), (it was trying to load the `jquery` relative path), but that got fixed when I used `makePromisClosure` function to close over the `sync_import` variable properly (weird that JS requires you to create that function to create closures)... However, the React example is using the exact same code path but failing.

assets/basics/node.js
JobJob: Yeah, I missed the recursive call to doImports before. I don't fully follow the code here, but I'll take your word for it that it makes sense.

src/widget.jl
JobJob: I didn't mean it that way. It was just when I saw the example code, I was like - how does that work? 

Anyway, I think this is prob fine and it could well change anyway if we move to global/non-widget based observable syncing.

README.md
JobJob: e.g. this one from this PR: 
> Dependencies which need to be loaded synchronously can be put in a Sync(["a.js", "b.js"]) wrapper. For examples dependencies=[Sync(["a.js", "b.js"]), "c.js"] will load b.js after a.js is loaded, but c.js in parallel with a.js and/or b.js.

Also could Sync be made not to require the array? e.g. 
`Sync("a.js", "b.js")` not `Sync(["a.js", "b.js"])`

shashi:
Got this thing to work by internally calling `SystemJS.config` when a dependency is named.

e.g.
```
deps = Sync(["react" => "https://cdnjs.cloudflare.com/ajax/libs/react/16.0.0/umd/react.development.js",
             "https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.0.0/umd/react-dom.development.js",
             "https://cdn.jsdelivr.net/npm/create-react-class@15.6.2/create-react-class.js"])
```

works. I guess naming some dependencies is unavoidable because of the way JS libraries are.

shashi:
Here's the extremely compact version:

```julia
using WebIO

deps = Sync(["react" => "https://cdnjs.cloudflare.com/ajax/libs/react/16.0.0/umd/react.development.js",
        ["https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.0.0/umd/react-dom.development.js",
         "https://cdn.jsdelivr.net/npm/create-react-class@15.6.2/create-react-class.js"]])


widget = Widget(imports=deps)

onimport(widget, @js function(React, ReactDOM, createReactClass)
    @var helloComponent = createReactClass(d(
            render = () -> React.createElement("strong", nothing, "Hello, World!")
        )
    )
    ReactDOM.render(
        React.createElement(helloComponent, null, null),
        this.dom,
    )
end)

widget
```

PR #46, opened 2018-02-22
Hook into Jupyter and serve assets

shashi:
Allows InteractNext to finally work offline!


deps/jlstaticserve.py
```diff
@@ -0,0 +1,36 @@
...
+    def get(self, pkg, fpath):
+        dirs = self.julia_load_path
+        for d in dirs:
+            dirpath = os.path.join(d, pkg, "assets")
+            fullpath = os.path.join(dirpath, fpath)
+            if os.path.isdir(dirpath) and os.path.isfile(fullpath):
+                self.root = dirpath
+                StaticFileHandler.get(self, fpath)
+                return
+        self.finish("Package gotthilla")
```
rohitvarkey: Hahaha! This is awesome! 😄

shashi: Lol, overlooked that when committing.

shashi:
This now works inside JupyterLab!

screenshot from 2018-02-23 06-46-15

shashi:
However requires the user to run this from `assets` dir:

```
jupyter labextension install webio
jupyter labextension enable webio/jupyterlab_entry
```
After editing any JS files, run

```
jupyter lab build
```

src/providers/ijulia_setup.jl
```diff
@@ -4,7 +4,7 @@ using IJulia.CommManager
 using WebIO
 
 function script(f)
-    display(HTML("<script>"*readstring(f)*"</script>"))
+    display(HTML("<script src='/pkg/WebIO/webio/bundle.js'></script>"))
```
sglyon: First off — amazing Worse to get ijulia asset server and and JupyterLab working in the same day!

In this function I don;t think it uses the `f` argument anymore, which can be confusing.

Issue #62, opened 2018-03-09
Way to include google fonts


djsegal:
Is there a way to add google fonts to a Scope through imports?

```
<link href="https://fonts.googleapis.com/css?family=Press+Start+2P" rel="stylesheet">
```

shashi:
`Scope(imports=["https://fonts.googleapis.com/css?family=Press+Start+2P"])` should work?

djsegal:
I get the following error :'(

<img width="1680" alt="screen shot 2018-03-09 at 12 42 44 pm" src="https://user-images.githubusercontent.com/3156114/37221306-777d4d0a-2397-11e8-9ae0-acb0f04723ff.png">

PR #63, opened 2018-03-10
Refactor import to get google fonts working (/css)

djsegal:
resolves:

+ #62


djsegal:
@shashi, how do you feel about this?

kind of seems like a stretch that anyone else (not google) would use something like:
+ `/css?blahblahblah`

shashi:
Looks good! Thanks!

Issue #72, opened 2018-03-20
Get Binder working across all browsers


djsegal:
Continuation of issue https://github.com/JuliaGizmos/WebIO.jl/issues/68 // was just too cluttered

Scenario now is:

> I have a binder that serves assets, but can't correctly find them

-------

i.e.

website looks for assets at:
+ http://example.com/hub/pkg/WebIO/webio/dist/bundle.js?_=1521526313580

but if you hunt around with chrome, you find they're actually at:
+ http://example.com/user/djsegal-webio-binder-o0v60382/pkg/WebIO/webio/dist/bundle.js?_=1521526313580

and this is because the user is at the jupyter notebook: 
+ http://example.com/user/djsegal-webio-binder-o0v60382/notebooks/julia.ipynb

-----

What part of the code do you think is responsible for this?

djsegal:
@shashi, is it related to: 

https://github.com/JuliaGizmos/WebIO.jl/blob/4807fc33dd46a115183d507894a4bbe8d5e6b37e/deps/jlstaticserve.py#L37-L39

Just trying to get my bearings before I setup a development environment for webio

shashi:
Unfortunately I think this is a `<base>` setting being done outside of WebIO https://www.w3schools.com/tags/tag_base.asp

I'm all ears if there are workarounds!

djsegal:
I've started with some workarounds:

+ https://github.com/djsegal/WebIO.jl/commit/102359a7ab201d17f4cd04d4f03fd81771c47cec

Another might be something a la:

+ https://discourse.julialang.org/t/how-to-get-url-of-current-jupyter-notebook/9851

shashi:
Can we close this now?

djsegal:
Closing :)

PR #75, opened 2018-03-21
Get WebIO working on Binder

djsegal:
attempts to resolve: 

+ https://github.com/JuliaGizmos/WebIO.jl/issues/72

### DIFF
--- a/src/providers/ijulia_setup.jl
+++ b/src/providers/ijulia_setup.jl
@@ -20,7 +20,11 @@ function main()
         <script>
             var curMatch =
                 window.location.href
-                .match(/(.*)\\/notebooks\\/(.*)\\.ipynb/);
+                .match(/(.*)\\/notebooks\\/.*\\.ipynb/);
+
+            curMatch = curMatch ||
+                window.location.href
+                .match(/(.*)\\/apps\\/.*\\.ipynb/);
 
             if ( curMatch ) {
                 \$('head').append('<base href="' + curMatch[1] + '/">');
--- a/src/providers/ijulia_setup.jl
+++ b/src/providers/ijulia_setup.jl
@@ -17,7 +17,7 @@ end
 
 function main()
     display(HTML("""
-        <script>
+        <script class='js-remove-script'>
             var curMatch =
                 window.location.href
                 .match(/(.*)\\/notebooks\\/.*\\.ipynb/);
@@ -32,8 +32,10 @@ function main()
         </script>
     """))
 
-    display(HTML("<script src='pkg/WebIO/webio/dist/bundle.js'></script>"))
-    display(HTML("<script src='pkg/WebIO/providers/ijulia_setup.js'></script>"))
+    display(HTML("<script class='js-remove-script' src='pkg/WebIO/webio/dist/bundle.js'></script>"))
+    display(HTML("<script class='js-remove-script' src='pkg/WebIO/providers/ijulia_setup.js'></script>"))
+
+    display(HTML("<script class='js-remove-script'>\$('.js-remove-script').closest('.output_area').remove();</script>"))
 
     comm = Comm(:webio_comm)
     conn = IJuliaConnection(comm)
--- a/src/providers/ijulia_setup.jl
+++ b/src/providers/ijulia_setup.jl
@@ -16,8 +16,20 @@ function WebIO.register_renderable(T::Type)
 end
 
 function main()
-    display(HTML("<script src='/pkg/WebIO/webio/dist/bundle.js'></script>"))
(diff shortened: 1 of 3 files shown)


src/providers/ijulia_setup.jl
```diff
@@ -16,15 +16,28 @@ function WebIO.register_renderable(T::Type)
...
-    display(HTML("<script src='/pkg/WebIO/webio/dist/bundle.js'></script>"))
-    display(HTML("<script src='/pkg/WebIO/providers/ijulia_setup.js'></script>"))
+    display(HTML("""
+        <script>
+            var curMatch =
+                window.location.href
+                .match(/(.*)\\/notebooks\\/(.*)\\.ipynb/);
+
+            if ( curMatch ) {
+                \$('head').append('<base href="' + curMatch[1] + '/">');
```
shashi: is `curMatch[1]` correct?

djsegal:
Shoot. Youre right. Ill change it to zero

Its hard going between js and julia

djsegal:
My bad. I lied. Its because i parened the main url leading to /notebooks/

1 is right. Could remove the parens

src/providers/ijulia_setup.jl
djsegal: I don't know if it is as correct as much as it works.

The original idea at the time of writing it was that I was trying to parse out the `hostname` too.

----

It does work, though.

Alternatively, we just change:

```julia
                .match(/(.*)\\/notebooks\\/(.*)\\.ipynb/);

            if ( curMatch ) {
                \$('head').append('<base href="' + curMatch[1] + '/">');
```

to

```julia
                .match(/.*\\/notebooks\\/(.*)\\.ipynb/);

            if ( curMatch ) {
                \$('head').append('<base href="' + curMatch[0] + '/">');
```

shashi: It's confusing to me why you want to extract the name of the notebook.

I thought you wanted to set the base url to everything leading up to the `/notebooks/`. I'm probably misunderstanding something.

djsegal: what i said was completely wrong.

+ `curMatch[0]` = the full url (not a partial of the url)

// don't accept this PR for another day or so

djsegal:
@shashi this is all done at this point.

hope i cleared up that one issue. let me know if you have any more thoughts. cheers!

src/providers/ijulia_setup.jl
shashi: LOL why not?

djsegal: Sorry. That warning about not accepting PR was pre some other changes.

API for match is: 

+ https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

-------

Quick example for the JS console:

```javascript
> tmpMatch = window.location.href.match(/(.*)\/JuliaGizmos/)
["https://github.com/JuliaGizmos", "https://github.com", index: 0, input: "https://github.com/JuliaGizmos/WebIO.jl/pull/75", groups: undefined]
```

and then

```javascript
> tmpMatch[0]
"https://github.com/JuliaGizmos"
> tmpMatch[1]
"https://github.com"
> tmpMatch[2]
undefined
```

src/providers/ijulia_setup.jl
```diff
@@ -32,8 +32,10 @@ function main()
         </script>
     """))
 
-    display(HTML("<script src='pkg/WebIO/webio/dist/bundle.js'></script>"))
-    display(HTML("<script src='pkg/WebIO/providers/ijulia_setup.js'></script>"))
+    display(HTML("<script class='js-remove-script' src='pkg/WebIO/webio/dist/bundle.js'></script>"))
+    display(HTML("<script class='js-remove-script' src='pkg/WebIO/providers/ijulia_setup.js'></script>"))
+
+    display(HTML("<script class='js-remove-script'>\$('.js-remove-script').closest('.output_area').remove();</script>"))
```
shashi: This will cause reloading the page to lose the JS....

djsegal: Very interesting... I'll just do a padding = 0 trick then!

djsegal: Sorry for trying to slip that in

src/providers/ijulia_setup.jl
shashi: Ah, nice! Looks good then. Sorry for not being schooled about how match works. If you revert the scrub output divs commit, I'll hit merge. Appreciate this useful feature!

src/providers/ijulia_setup.jl
djsegal: Done!

PR #77, opened 2018-03-22
Enable messages to be sent to many web connections

### LINKED ISSUE #71: Pushing updates to multiple web clients
I was investigating whether it's possible to have multiple web views receiving the same updates from Julia. It seems like, at least from the basic demos, this sort of works. That is, pushing an update from Julia to Javascript works, but only one web view receives the update, and which one appears to be random. 

To reproduce, I started with the simple clock demo from the readme:

```julia
w = Scope()
obs = Observable(w, "clock-value", "")

timestr() = Dates.format(now(), "HH:MM:SS")

# update timestamp every second
@async while true
    sleep(1)
    obs[] = timestr()
end

# on every update to `obs`, replace the text content of #clock
onjs(obs, @js val -> begin
    @var clock = this.dom.querySelector("#clock")
    clock.textContent = val
end)

node = w(
  dom"div#clock"(
    timestr(),
  ),
)
```

I let IJulia display that final `node` variable, which results in the ticking clock. I then set up a Mux server to serve the same node:

```julia
using Mux
function myapp(req) # an "App" takes a request, returns the output
    node
end

webio_serve(page("/", myapp))
```

And then I opened `localhost:8000` in a browser.

The result is that both clocks keep ticking, but only one of them ticks in a given second, so each individual clock only updates every few seconds. 

Is this expected? Do we want to make this work? Or am I just doing it wrong? In MeshCat.jl I manage this by deliberately sending every update to all the connected websockets, and I suspect I could do something similar with WebIO, but only if that's actually desirable. 

Thanks!

rdeits:
This is a rough proposal to fix #71. It replaces the `Scope`'s `outbox` field with a new struct, `ConnectionPool` that holds both the outbox of messages and a set of connections to send those messages to. On construction, the connection pool starts a task that will block until a connection is made, and then will send each message the outbox receives to every open connection. 

Anyway, please let me know if this seems like the right solution.

### DIFF
--- a/src/providers/blink_setup.jl
+++ b/src/providers/blink_setup.jl
@@ -26,6 +26,9 @@ function Base.send(b::BlinkConnection, data)
     Blink.msg(b.page, Dict(:type=>"webio", :data=>data))
 end
 
+# TODO: I'm not sure what to check here
+Base.isopen(b::BlinkConnection) = true
+
 function WebIO.register_renderable(T::Type)
     Blink.body!(p::Union{Window, Page}, x::T) =
         Blink.body!(p, WebIO.render(x))
--- a/src/providers/blink_setup.jl
+++ b/src/providers/blink_setup.jl
@@ -26,8 +26,7 @@ function Base.send(b::BlinkConnection, data)
     Blink.msg(b.page, Dict(:type=>"webio", :data=>data))
 end
 
-# TODO: I'm not sure what to check here
-Base.isopen(b::BlinkConnection) = true
+Base.isopen(b::BlinkConnection) = Blink.active(b.page)
 
 function WebIO.register_renderable(T::Type)
     Blink.body!(p::Union{Window, Page}, x::T) =
--- a/src/providers/ijulia_setup.jl
+++ b/src/providers/ijulia_setup.jl
@@ -11,6 +11,9 @@ function Base.send(c::IJuliaConnection, data)
     send_comm(c.comm, data)
 end
 
+# TODO: is this the right thing to check?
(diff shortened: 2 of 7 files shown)


src/providers/blink_setup.jl
```diff
@@ -26,6 +26,9 @@ function Base.send(b::BlinkConnection, data)
     Blink.msg(b.page, Dict(:type=>"webio", :data=>data))
 end
 
+# TODO: I'm not sure what to check here
+Base.isopen(b::BlinkConnection) = true
```
shashi: Maybe that the Blink window is open?

shashi:
Very nice! :100:

Are you still working on this patch or is it ready to merge?

rdeits:
I'll just fix those two remaining todos and then it should be good.

src/providers/blink_setup.jl
rdeits: Ok, done. I found `Blink.active` which seems to do the right thing.

src/providers/ijulia_setup.jl
```diff
@@ -11,6 +11,9 @@ function Base.send(c::IJuliaConnection, data)
     send_comm(c.comm, data)
 end
 
+# TODO: is this the right thing to check?
+Base.isopen(c::IJuliaConnection) = haskey(IJulia.CommManager.comms, c.comm.id)
```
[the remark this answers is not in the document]
rdeits: Ok, sounds good

rdeits:
Ok, should be good to go once CI passes.

shashi:
Thanks!!

PR #78, opened 2018-03-22
use homedir instead of accessing ENV directly

### LINKED ISSUE #67: No home
https://github.com/JuliaGizmos/WebIO.jl/blob/11389b75f0631d01dc3bcc924d8facc0e253309a/deps/build.jl#L6

This line assumes ENV["HOME"] is defined - which it isn't?! At least on windows it isn't ;)

SimonDanisch:
fixes #67


shashi:
Thank you! :-)

PR #83, opened 2018-03-27
fix order of operations in process_messages

rdeits:
Previously, if you did the following:

1. add connection A
2. add connection B
3. send message 1
4. send message 2

then connection A would receive messages 1 and 2, but connection B would only receive 2. That's because the `process_messages()` task would advance past `ensure_connection` when the first connection was found, which meant that connection 2 would not be found until the second message. 

Flipping the order in which we wait for connections and messages fixes the issue, and I've added some tests to make sure.

### DIFF
--- a/src/scope.jl
+++ b/src/scope.jl
@@ -57,8 +57,8 @@ end
 
 function process_messages(pool::ConnectionPool)
     while true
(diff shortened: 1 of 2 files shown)


shashi:
:+1:

PR #84, opened 2018-03-28
WIP: IFrame wrapper

shashi:
Can be used to fight pre-existing stylesheets and event handlers.

cc @rdeits 

I'm now trying to get the height to be the same as content (doesn't work when content < 150px). ~Another issue is observables don't sync between the parent and the frame (e.g. slider)~

### DIFF
--- a/assets/webio/dist/bundle.js
+++ b/assets/webio/dist/bundle.js
@@ -106,9 +106,6 @@ function arrays_and_equal(arr1, arr2){
     return is_array(arr1) && is_array(arr2) && arrays_equal(arr1, arr2)
 }
 
-var scopes = {};
-var obsscopes = {};
-
 function makeScope(id, data, sendCallback, dom, handlers, observables)
 {
     var scope = {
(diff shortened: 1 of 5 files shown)


rdeits:
This is awesome, thank you! I'll try it out right away.

rdeits:
I'm getting syntax errors in Firefox from the `javascript:void()` `src` attribute. It seems to work anyway, but the errors go away if I change that to `javascript:void(0)`

shashi:
Okay we should use that then!

rdeits:
Ok, this is working great for rendering my content. Thank you! The only missing piece is the ability to send observables from Julia into the iframe. I'm still trying to figure out how that would work.

shashi:
You mean just using Observables in Scope that is wrapped in `iframe`?

Doesn't that work?

I can do

```julia
In[1]: s = slider(1:10)
In[2]: iframe(s)
In[3]: WebIO.render(observe(s))
```
and both sliders update the observable...

rdeits:
Is that `slider()` from InteractNext.jl? I'll give it a shot. 

Sorry, what I meant was that I tried pushing updates from Julia to JS using the clock demo from the readme:

```julia
w = Scope()
obs = Observable(w, "clock-value", "")

timestr() = Dates.format(now(), "HH:MM:SS")

# update timestamp every second
@async while true
    sleep(1)
    obs[] = timestr()
end

# on every update to `obs`, replace the text content of #clock
onjs(obs, @js val -> begin
    console.log(val);
    @var clock = this.dom.querySelector("#clock")
    clock.textContent = val
end)

WebIO.iframe(w(
  dom"div#clock"(
    timestr(),
  ),
))
```

but the resulting clock doesn't update. Is that expected?

shashi:
Oh I see. No we should definitely fix that before merging this. I'll have a look.

rdeits:
Ok, thank you!

shashi:
Just pushed the required fixes. Clock works, and sliders also sync :-)

rdeits:
Yup, confirmed that it works for MeshCat.jl too!

PR #89, opened 2018-03-29
test `onjs()` and round-trip communication from Julia to JS

rdeits:


test/blink-tests.jl
```diff
@@ -1,12 +1,30 @@
...
+    c = Channel{Any}(1)
+    @async begin
+        put!(c, f())
+    end
+    @async begin
+        sleep(timeout)
+        put!(c, nothing)
+    end
+    take!(c)
+end
```
shashi: This is cool. There is a `sleep(5)` somewhere in tests, maybe we can use this to get rid of that.

PR #99, opened 2018-05-03
Fix JS imports on JuliaBox

### LINKED ISSUE #96: WebIO imports don't work on JuliaBox
For example, try running the following in a `next.juliabox.com` notebook:

```julia
using WebIO

scope = WebIO.Scope(
    imports=[
        "pkg/WebIO/webio/dist/bundle.js"
    ]
)
```

This results in two errors in the javascript console:

```
Failed to load https://auth.juliacomputing.io/dex/auth?response_type=code&client_id=dev-juliabox&state=a1c0437e48a839a5003047b1aa2620a4&redirect_uri=https%3A%2F%2Fnext.juliabox.com%2Fauth%2Flogin&nonce=4e339121edd0884f63d85171dd1343f2&scope=openid%20email%20profile%20offline_access: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://next.juliabox.com' is therefore not allowed access. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
bundle.js:2571 Uncaught (in promise) Error: Failed to fetch
  Instantiating https://next.juliabox.com/notebook/pkg/WebIO/webio/dist/bundle.js
  Loading pkg/WebIO/webio/dist/bundle.js
```

The URL being referenced definitely exists, which we can demonstrate by including it in a regular `script` tag with: `display(HTML("<script src='pkg/WebIO/webio/dist/bundle.js'></script>"))`. But it looks like the way WebIO loads imports is failing to satisfy CORS. 

Looking into this, it appears that WebIO is using SystemJS to load JS dependencies. It should be possible to tell SystemJS to use `script` tags instead of XHR with something like: 

```javascript
System.config({
    meta: {
        '*': {
            scriptLoad: true
        }
    }
})
```

but this seems to have no effect. 

Alternatively, this thread suggests that `authorization: true` might help, and that does change the error output slightly: 

```js
System.config({
    meta: {
        '*': {
            authorization: true,
        }
    }
});
SystemJS.import('pkg/WebIO/webio/dist/bundle.js');
```

```
Promise {<rejected>: Error: Failed to fetch
  Instantiating https://next.juliabox.com/notebook/pkg/WebIO/webio/dist/bundl…}
bundle.js:2571 Uncaught (in promise) Error: Failed to fetch
  Instantiating https://next.juliabox.com/notebook/pkg/WebIO/webio/dist/bundle.js
  Loading pkg/WebIO/webio/dist/bundle.js
```

### LINKED ISSUE #98: Incorrect webcomponents URL on JuliaBox in iframe
On JuliaBox, WebIO tries to load the webcomponents module from:

`https://juliabox.com/pkg/WebIO/webio/dist/webcomponents-lite.min.js` 

but instead, it needs to be:

`https://juliabox.com/notebook/pkg/WebIO/webio/dist/webcomponents-lite.min.js`

It's probably just another absolute vs. relative path thing. I'll see if I can fix it.

### LINKED ISSUE #101: Regenerating javascript bundle breaks scopes with imports

```julia
using WebIO
WebIO.devsetup()
WebIO.bundlejs()
```

```js
```

rdeits:
Supersedes #97, fixes #98, fixes #96, fixes #101 

Edit: now rebased on #103


rdeits:
Hm, there's something wrong here. When I try to test MeshCat.jl over WebIO locally, I get `Error: cannot find module "vm"` in the javascript console. I suspect something has gone wrong in the bundling process.

djsegal:
I'll try to check this out tonight! Are there any reader's digest tips about what to look out for?

rdeits:
@djsegal you might want to wait to test this out. There's something wrong with the bundling process, at least on my machine (see #101 ). It's happening for me on `master` and also on this PR, so it will make testing harder.

rdeits:
Safest thing to do seems to be to only set `{authorization: true}` for local `pkg` imports.

shashi:
That sounds OK.

shashi:
from this discussion https://github.com/systemjs/systemjs/issues/1150 it seems like `{authorization: true}` sends the cookies in the XHR request. It probably sends the same to external CDN servers as well, which they are at full liberty to refuse to respond to.

I'm alarmed now by the extra 7k lines in this diff!! Let's try and fix that before merging.... I can try building it on my local setup when you're done editing.

shashi:
I see most of it is package-lock.json... Did you figure out if we can go back to webpack 0.21?

rdeits:
Ok, sounds good. The extra 7k lines are all because I checked in `package-lock.json` (which is [recommended](https://stackoverflow.com/a/44210813/641846)). Actually, the changes to the `js` bundles are probably bigger but have fewer line changes because `webpack` squishes everything into a single line. 

But if you want to minimize the impact of this PR, I can remove `package-lock.json` from this PR and we can discuss committing it later.

rdeits:
I get the `Critical dependency: require function is used in a way in which dependencies cannot be statically extracted` warning for systemjs 0.21.3, 0.21.2, and 0.21.1. It looks like 0.21.0 is OK, though. I suspect that was the version available when the bundle was last successfully built (it was the latest version from mid-January through the end of March 2018). I'll update the `package.json` to use that version.

rdeits:
https://github.com/systemjs/systemjs/issues/1817

shashi:
I'm ok to merge this and wait for resolution of the SystemJS issue. Then we can try and upgrade.

Thanks a million for looking into these issues. This is great!

PR #102, opened 2018-05-03
Fix weird bundling issue with systemjs

### LINKED ISSUE #101: Regenerating javascript bundle breaks scopes with imports

rdeits:
Fixes #101 

The important change here seems to be restricting systemjs to 0.20 instead of 0.21. Otherwise running the `bundlejs()` step produces the following warning: 

```
WARNING in ./node_modules/systemjs/dist/system.js
    4:36270-36272 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
     @ ./node_modules/systemjs/dist/system.js
     @ ./node.js
     @ ./index.js
```

and we get the errors in #101. 

I'd love to find a better solution that doesn't require downgrading systemjs, but I'm not sure what that would be.


shashi:
It's really annoying that newer versions of systemjs breaks things that used to work.

So having fixed the versions, can we be sure that everyone compiling the JS will generate small diffs?

shashi:
It seems

```
     @ ./node.js
     @ ./index.js
```

Probably hints that our code is doing something illegal.

rdeits:
I think that the best way to ensure future compatibility would be to check in the `package-lock.json` file, which will ensure exactly the same versions of everything going forward. I'm not entirely sure how that will work with the two separate node package installs (in `deps` and `assets`). What's the reason for having two entirely separate `node_modules` directories? They both contain copies of webpack and all its dependencies, so there's a lot of duplication there.

shashi:
Having 2 different package.json was the easiest way to build in both `assets` and `deps`. I am open to unifying them. Any JS stuff can probably completely moved to `assets` without making the name nonsensical.

rdeits:
Closing in favor of #103

Issue #94, opened 2018-04-17
Vertx setup issue


epaaso:
Hi!
When running `WebIO.bundlejs()` I get this:

```
INFO: Installing NodeJS v0.4.0
INFO: Building NodeJS

up to date in 1.948s

npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

(node:5115) DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead

Hash: 4c3a3033c476ecba64c1e2727850f60988ff06d0
Version: webpack 4.5.0
Child
    Hash: 4c3a3033c476ecba64c1
    Time: 360ms
    Built at: 2018-4-17 11:55:19
        Asset     Size  Chunks             Chunk Names
    bundle.js  128 KiB       0  [emitted]  bundle
    Entrypoint bundle = bundle.js
       [0] (webpack)/buildin/global.js 509 bytes {0} [built]
       [1] ./webio.js 6.47 KiB {0} [built]
       [9] ./node.js 10.4 KiB {0} [built]
      [12] ./unsafe_script.js 1.83 KiB {0} [built]
      [13] ./index.js 90 bytes {0} [built]
        + 9 hidden modules
Child
    Hash: e2727850f60988ff06d0
    Time: 305ms
    Built at: 2018-4-17 11:55:19
                        Asset      Size  Chunks             Chunk Names
    webcomponents-lite.min.js  83.1 KiB       0  [emitted]  webcomponents-lite.min
    Entrypoint webcomponents-lite.min = webcomponents-lite.min.js
       [1] (webpack)/buildin/global.js 509 bytes {0} [built]
        + 2 hidden modules
    
    WARNING in ./node_modules/@webcomponents/webcomponentsjs/webcomponents-lite.js
    Module not found: Error: Can't resolve 'vertx' in '/home/ernesto/.julia/v0.6/WebIO/assets/webio/node_modules/@webcomponents/webcomponentsjs'
     @ ./node_modules/@webcomponents/webcomponentsjs/webcomponents-lite.js 28:263-279

INFO: Removing NodeJS v0.4.0
```

[...]

Issue #79, opened 2018-03-23
setup issue


AshtonSBradley:
I am trying to install this to use `InteractNext`. When setting up:

```julia
julia> using WebIO

julia> WebIO.devsetup()
INFO: No packages to install, update or remove
npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.
npm WARN deprecated babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update! 
npm WARN bootstrap@4.0.0 requires a peer of popper.js@^1.12.9 but none is installed. You must install peer dependencies yourself.
npm WARN abradley@1.0.0 No description
npm WARN abradley@1.0.0 No repository field.

+ webpack-cli@2.0.13
+ webpack@4.2.0
updated 2 packages in 22.707s
INFO: Removing NodeJS v0.4.0

julia> WebIO.bundlejs(watch=false)
INFO: Installing NodeJS v0.4.0
INFO: Building NodeJS
up to date in 2.58s
module.js:549
    throw err;
    ^

Error: Cannot find module '/Users/abradley/.julia/v0.6/WebIO/deps/node_modules/webpack-cli/bin/webpack.js'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:188:16)
    at bootstrap_node.js:609:3
ERROR: LoadError: failed process: Process(`/Users/abradley/.julia/v0.6/NodeJS/deps/bin/node-v8.10.0-darwin-x64/bin/node ../../deps/node_modules/webpack-cli/bin/webpack.js`, ProcessExited(1)) [1]
Stacktrace:
 [1] pipeline_error(::Base.Process) at ./process.jl:682
 [2] run(::Cmd) at ./process.jl:651
 [3] include_from_node1(::String) at ./loading.jl:576
 [4] include(::String) at /Applications/Julia-0.6.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [5] #bundlejs#42(::Bool, ::Function) at /Users/abradley/.julia/v0.6/WebIO/src/devsetup.jl:12
 [6] (::WebIO.#kw##bundlejs)(::Array{Any,1}, ::WebIO.#bundlejs) at ./<missing>:0
[...]

PR #103, opened 2018-05-04
Remove devsetup and clean up bundling

### LINKED ISSUE #101: Regenerating javascript bundle breaks scopes with imports

rdeits:
Replaces #102, fixes #101 

I was able to get rid of the `devsetup()` step entirely, and get rid of its associated `node_modules` folder in `deps`. Since we were already doing a `npm install` in the `bundlejs()` step, I don't think there's much reason for a separate `devsetup()`. 

I've tightened the version restrictions in `package.json` and checked in `package-lock.json` so we'll have a clear idea when package versions change. Unfortunately, this doesn't *guarantee* future compatibility, because `npm install` overwrites, rather than respecting, `package-lock.json`. It looks like making that guarantee will be possible with `npm 5.7.1` using the `npm ci` command: https://github.com/npm/npm/issues/18286#issuecomment-365130068

By the way, `bundlejs(watch=true)` still works after this change :slightly_smiling_face: 

Also, we should check if this helps with #79 and #94


shashi:
I'm not sure now if this still supersedes #99. Let me know and I'll merge the right PR.

Thanks!

rdeits:
Yeah, sorry, I tried to keep the changes here separate from #99 but it all got kind of mixed together in the end. I'll close this so we can focus on a single set of changes.

Issue #107, opened 2018-05-07
Scopes with imports don't work in Juno or Blink


rdeits:
(split off from #105)

Javascript imports inside Scopes don't load properly in Juno. For example, the `p5.js` demo from the readme fails with:

```
Fetch API cannot load http://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js. Redirect from 'http://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js' to 'https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://127.0.0.1:3230' is therefore not allowed access. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
2plots.coffee [sm]:39 Uncaught (in promise) Error: Failed to fetch
  Instantiating http://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js
  Loading //cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js
```

`/pkg` imports also fail, but that's not surprising since the Jupyter server isn't there to serve those requests. 

This happens both before and after the systemjs changes in #99

rdeits:
I've checked back to v0.1.4, and I can't find any versions where `examples/p5.jl` works in Blink. I still don't really understand the scope import machinery outside of IJulia, so I'm not sure how helpful I can be here.

rdeits:
Hm. Everything is working now. I can run the p5.js demo app in both Blink and Juno with no issues. 

I wonder if perhaps I was just having intermittent internet access problems when I was testing this?

shashi:
Little known trick: you can call `tools(w)` to open developer tools/js console on blink window `w`.

rdeits:
Cool! I'd just been hacking the `main.js` in Blink.jl to open the dev tools, but that's much nicer.

PR #110, opened 2018-05-09
Add a test for scope imports in Blink

rdeits:
This PR adds tests for scope imports with the following URL types:

* `/pkg/Foo/...`
* `pkg/Foo/...`
* `//public/url/...`
* `http://public/url/...`

~~Currently only the first one actually works on Blink (see #107 ), so we should fix those issues before merging.~~

Edit: all 4 URLs are working now! 

In the long run, we can change the public URL used here to the permanent github URL of the `trivial_import.js` file, but since it's not committed yet I'm currently just serving it from my personal public folder. That's easy to change later.

### DIFF
--- a/assets/webio/test/trivial_import.js
+++ b/assets/webio/test/trivial_import.js
@@ -1,0 +1,6 @@
+// A trivial importable module that simply produces "ok".
+// Used for testing the WebIO Scope imports machinery.
+(function() {
+	x = "ok";
+	return {x: x}
+})()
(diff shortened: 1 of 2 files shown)


rdeits:
I just checked, and all 4 URL versions do work when served with `webio_serve` and Mux, so this is definitely something specific to Blink.

rdeits:
Aha! All four imports actually work if I create a new `Blink.Window()` each time. They also worked in Mux because I was creating a new Mux server for each separate URL. 

What's going on is that `SystemJS.import("foo.js")` only performs the import if `foo.js` has not already been loaded. As a result, the `WebIO.onimport` callback is only passed the imported module the *first* time a given javascript file is imported. Any subsequent imports of the same javascript file will result in `onimport` being called with an empty object. Since that's how SystemJS works (and apparently how the browser fetching works internally), maybe this is just working as intended. 

I've fixed the tests included here by just creating 4 separate javascript files to import, to ensure that `onimport` actually fires.

PR #113, opened 2018-05-14
fix duplicate webio jupyter notebook config entries

### LINKED ISSUE #112: Build process creates duplicate jupyter webio configs
Related to #100 but not the cause of that issue. 

The current build script looks for `Regex("\n" * BEGIN_MARKER * ".*" * END_MARKER, "s")`, which will fail if the webio config is already present but does not have a newline before it. That's exactly the situation that occurs if your config is initially empty, resulting in a jupyter notebook config with exactly two webio config blocks. For example, here's mine: 

```python
###JULIA-WEBIO-CONFIG-BEGIN
import sys, os
if os.path.isfile("/home/rdeits/.julia/v0.6/WebIO/deps/jlstaticserve.py"):
    sys.path.append("/home/rdeits/.julia/v0.6/WebIO/deps")
    c = get_config()
    c.NotebookApp.nbserver_extensions = {
        "jlstaticserve": True
    }
else:
    print("WebIO config in ~/.jupyter/jupyter_notebook_config.py but WebIO plugin not found")
###JULIA-WEBIO-CONFIG-END
###JULIA-WEBIO-CONFIG-BEGIN
import sys, os
if os.path.isfile("/home/rdeits/locomotion/explorations/learning-mpc-2/packages/v0.6/WebIO/deps/jlstaticserve.py"):
    sys.path.append("/home/rdeits/locomotion/explorations/learning-mpc-2/packages/v0.6/WebIO/deps")
    c = get_config()
    c.NotebookApp.nbserver_extensions = {
        "jlstaticserve": True
    }
else:
    print("WebIO config in ~/.jupyter/jupyter_notebook_config.py but WebIO plugin not found")
###JULIA-WEBIO-CONFIG-END
```

The first config never matches the pattern, so it can never be removed. 

This comes about because the config string which we add starts with:

```
"""
$BEGIN_MARKER
...
```

but the leading newline in a `"""` block is ignored in Julia.

rdeits:
fixes #112

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -12,10 +12,11 @@ function install_ijulia_config()
     end
 
     # remove previous config
-    config_str = replace(config_str, Regex("\n" * BEGIN_MARKER * ".*" * END_MARKER, "s"), "")
+    config_str = replace(config_str, Regex("\n?" * BEGIN_MARKER * ".*" * END_MARKER * "\n?", "s"), "")
 
     loadpath = JSON.json(vcat(Pkg.dir(), LOAD_PATH))
     config_str *= """
+
     $BEGIN_MARKER
     import sys, os
     if os.path.isfile("$(joinpath(dirname(@__FILE__), "jlstaticserve.py"))"):


shashi:
That's a nice bug! Thanks for catching it!

Issue #100, opened 2018-05-03
Assets are loaded from wrong WebIO install in JuliaBox


rdeits:
When WebIO is built, we do: https://github.com/JuliaGizmos/WebIO.jl/blob/204f58093942f3994c3ebea5b17eb333ededdad6/deps/build.jl#L17

However, JuliaBox actually has two copies of WebIO. There's one installed systemwide in:

```
/home/jrun/.julia/v0.6/WebIO
```

and one installed as part of the user package bundle in:

```
/mnt/juliabox/.julia/v0.6/WebIO
```

*both* of these folders are in `LOAD_PATH`, and the first one is also `Pkg.dir()` at the time that the WebIO plugin is built and installed. 

When you run `using WebIO`, you get the user-package from `/mnt/juliabox`, but when you try to import a javascript asset, WebIO's jupyter extension first looks in the system packages in `/home/jrun`. This results in a weird situation where WebIO's julia and javascript assets can come from different copies of the package (and potentially even different versions). This problem will also occur for any other package which is part of the default JuliaBox systemwide installation.

rdeits:
Yeah, I think matching Julia's priority is totally the right thing to do. We can do that as a general improvement, but I don't know if it will help the situation on JuliaBox because *both* package directories are also in LOAD_PATH when the package is built, and the systemwide one comes first.

rdeits:
This is becoming more of a problem for me in other contexts, and I'm afraid that just changing the preference order isn't going to resolve the situation. The particular problem I'm having is that I do a lot of work in separate Julia environments (with separate package directories and load paths), but WebIO can't pick up assets from anywhere but the default package directory and whatever happened to be on LOAD_PATH whenever it was first installed. 

For an example that doesn't involve JuliaBox at all, imagine: 

1. We install WebIO in the default Julia package directory `~/.julia/v0.6`
2. WebIO installs its jupyter extension to look for assets in `~/.julia/v0.6`
3. We then create a new julia package directory in `/elsewhere` and install some packages into it
4. We run a WebIO app using the julia packages from `/elsewhere` and try to load a `pkg` URL, but we end up getting assets from `~/.julia/v0.6` instead of `/elsewhere`. 

Problems like this seem inevitable given that WebIO is trying to create a single global list of package asset folders. And it seems likely that this will get worse with pkg3, which will make it easier to have nicely separated projects with independent packages. 

[...]

PR #121, opened 2018-05-18
make href regexes non-greedy

### LINKED ISSUE #120: Cannot use WebIO in IJulia inside a folder called `notebooks`
Yeah, this one was a bit of a stumper. 

If you create a folder called `notebooks`, start `jupyter notebook` outside that folder, and then try to create a notebook inside the `notebooks` folder, then using WebIO will fail. This happens because the regex to detect hosts which use a `notebooks` prefix, isn't quite right. When inside a folder literally called `notebooks`, the URL is: 

```
http://localhost:8888/notebooks/notebooks/foo.ipynb 
```

and the regex is: `/(.*)\\/notebooks\\/.*\\.ipynb/`. Since regexes are greedy, rather than matching `http://localhost:8888/`, the regex keeps going and ends up matching `http://localhost:8888/notebooks/`. That results in the wrong base URL, so all the WebIO pkg URLs are broken.

Basically: 

Anyway, I'll see if I can fix this. 

@tkoolen, by the way, this is why the new MeshCat didn't work the first time I tried to show you (because we were trying out one of your demos which happened to live in a folder called `notebooks`).

rdeits:
Fixes #120 

Obviously, the solution to our regex problem is more regex :upside_down_face:

### DIFF
--- a/src/iframe.jl
+++ b/src/iframe.jl
@@ -2,7 +2,7 @@ export iframe
 
 function iframe(dom)
     str = stringmime("text/html", dom)
-    
+
     s = Scope()
     s.dom = Node(:div,
                  Node(:iframe, id="ifr", style=Dict("width"=>"100%"),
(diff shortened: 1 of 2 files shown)


rdeits:
Yeah, agreed. Actually, I've been wondering if the `pkg` URL thing is the right solution for the long term. Issues like https://github.com/JuliaGizmos/WebIO.jl/issues/100#issuecomment-388903408 seem like they're going to keep cropping up because the `pkg` URLs are always provided by a single, global Jupyter server, but every single kernel can potentially have a completely different set of packages and associated assets. There's just no way for the global server to always do the right thing. 

What if we sent assets over the existing IJulia comms channel? Is that possible? The Julia kernel always knows where its package directory is, so it would solve #100 and all of the issues related to `<base>` URLs completely.

PR #129, opened 2018-05-29
fix some deprecations for 0.7

rdeits:
This doesn't quite get WebIO working on 0.7 due to some issues with Requires, but it's an improvement.


shashi:
Nice! Do merge when it's ready.

PR #130, opened 2018-05-30
fix p5.jl example with Blink

rdeits:
The real fix is https://github.com/JunoLab/Blink.jl/pull/89 but at least this makes the example work properly.


shashi:
:+1:

PR #135, opened 2018-06-05
Fix #134 : remove Widget binding

### LINKED ISSUE #134: Remove deprecated Widget type
This package exports the deprecated `Widget` type as a synonymy for `Scope`. Can it be simply removed? I would like to reclaim the word `Widget` for InteractBase

piever:


shashi:
It's also exported I think. Not on my laptop right now.

piever:
It seems like `@deprecate_binding` also exports it, so removing it effectively unexports `Widget`

shashi:
sounds good!

PR #137, opened 2018-06-05
Add LaTeXString support.

korsbo:
A very simple addition which allows [LaTeXStrings](https://github.com/stevengj/LaTeXStrings.jl) to be rendered. I have tested this for Jupyter.


korsbo:
I may have been a little too hasty in submitting this. It does not work in Juno.

PR #138, opened 2018-06-06
Math.rand() -> Math.random() in the Readme

rdeits:
because that's the actual javascript function :wink:


shashi:
:sweat_smile: thanks!

PR #141, opened 2018-06-18
Register bundle.js and helpers via AssetRegistry in IJulia and Mux

shashi:
cc @ranjanan

### DIFF
--- a/src/providers/ijulia.jl
+++ b/src/providers/ijulia.jl
@@ -2,6 +2,7 @@
 
 using IJulia
 using IJulia.CommManager
+using AssetRegistry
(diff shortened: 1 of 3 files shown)


src/providers/ijulia.jl
```diff
@@ -24,6 +25,7 @@ function main()
         # copy of WebIO and IJulia. 
         return
     end
+    key = AssetRegistry.register(joinpath(@__DIR__, "..", "..", "assets"))
```
ranjanan: Maybe make this path a global const so that it can also be statically compilable?

PR #149, opened 2018-06-27
Don't prevent an update if the value doesn't change

### LINKED ISSUE #39: Allow updates with the same value
Right now the "counter" examples are broken since updating with the same value drops the update.

shashi:
This is necessary to get Knockout to work properly if I understand correctly. cc @piever 

This makes https://github.com/JuliaGizmos/Vue.jl/pull/10 necessary. We should release them both in tandem.

closes #39 against @JobJob's will?? :-P Let me know if you still think we really need this check here! My reasoning is [here](https://github.com/JuliaGizmos/Vue.jl/pull/10#issuecomment-399850249)

### DIFF
--- a/assets/webio/dist/bundle.js
+++ b/assets/webio/dist/bundle.js
@@ -124,7 +124,7 @@ function makeScope(id, data, sendCallback, dom, handlers, observables)
     }
     WebIO.scopes[id] = scope;
     if (observables){
-        Object.keys(observables).forEach(function setobsscope(name){
+        Object.keys(observables).forEach(function setobsscope(name) {
             var o = observables[name]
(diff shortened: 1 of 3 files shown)


piever:
I think Knockout can somehow be patched to workaround this (like in Vue.jl, by always making a copy of the whole object that's synced between Knockout and WebIO) but I somehow feel it's strange to have different semantics between WebIO `Observables` and Julia `Observables` (especially as there is no way to forcefully "trigger" an update right now).

JobJob:
I think all the arguments in favour of this seem sensible. 

I have a nagging feeling that there was a case that made me want to use the "don't update unless changed "semantic, but since I can't think of it, we should merge this.

Issue #148, opened 2018-06-27
Base URL woes


shashi:
Various systems want to serve Jupyter/Mux on non-root URLs. For example, binder runs it on `mybinder.org/foobar/notebooks/Foo.ipynb`, appmode runs it on `/appmode/Foo.ipynb`

We set up our server code to serve assets at `/assetserver/` but this will of course not work on binder because `mybinder.org/assetserver/shaHash/foo.js` will obviously not be served there, but it will be served on `mybinder.org/foobarmy_binder_id/assetserver/shaHash/foo.js`

As an alternative you can try `../assetserver` which doesn't work on Mux anymore but kinda works as long as your notebook is in the root folder in Jupyter.

@djsegal's [hack here](https://github.com/JuliaGizmos/WebIO.jl/blob/c5b22ab576e3b99d5576a249a019b2b08c20f642/src/providers/ijulia.jl#L30-L44) tries to fix this by setting a `<base>` tag, but this causes unexpected issues like #95 

When serving a Mux app on JuliaBox (a new feature), the URL is `/apps/sha1hash/` and this is the root URL where Mux is served so all asset URLs need to be relative to this. This is now a different set up.

I think it's reasonable to expect that someone setting up a WebIO backend knows what's the root URL to the Mux/Jupyter server. We process each URL used as import in a Scope via `lowerdeps`, we allows us to reliably prefix this root URL.

We need to figure out how the user can configure the root URL. The obvious choice is:

```julia
using WebIO
WebIO.set_baseurl("/mybinderURL")
```
But one small problem is we inject the WebIO JavaScript in Jupyter right on `using WebIO` which means we already loaded the wrong URLs. But it may be OK to print the required JS also on a call to `WebIO.set_baseurl`.

Now in something like binder or tmpnb if the root URL is randomly generated for each instance, you'd have to use JavaScript to figure out the current URL and then send that to Julia --- however you don't yet have convenient communication since we don't yet have WebIO working... In such cases the user would have to use relative paths like `../` as base URL and hope no one moves the notebook into a folder at a different level.

simonbyrne:
For the Jupyter issue, you could also allow setting the URL via an environment variable (say `JULIA_WEBIO_BASEURL`), then call `set_baseurl` at `__init__`?

shashi:
Great idea!

shashi:
I have added the change to #154

PR #153, opened 2018-06-29
Make Mux work on non-root routes

shashi:
After doing this change I realized that we assume Mux to be running on the root of the domain when we assume `/assetserver/hash` urls will work. But this is quite robust anyway.


shashi:
Turns out we need this behavior on JuliaBox for mux apps.

shashi:
This should also be addressed by #148

PR #158, opened 2018-07-22
demonstrate simple rendering of Observables

rdeits:
This works just fine in IJulia, but in Blink, I had to do: 

```julia
body!(Window(), WebIO.render(clock_obs))
```

instead of:

```julia
body!(Window(), clock_obs)
```

Doing `register_renderable(Observable)` didn't seem to change that. Maybe the `register_renderable` for Blink isn't working?

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,5 +406,5 @@ clock_obs = Observable(timestr())
     sleep(1)
     clock_obs[] = timestr()
 end
-WebIO.render(clock_obs)
+clock_obs
 ```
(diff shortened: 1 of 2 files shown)


rdeits:
Edit: I also had to manually call `WebIO.render()` when using Mux: 

```julia
julia> webio_serve(page("/", req -> WebIO.render(clock_obs)), 8001)
```

shashi:
Adding the methods to call render on any input to the Mux / Blink clients would involve pirating their existing behavior of not doing that...

PR #159, opened 2018-07-22
Proposal: new `node()` vararg constructor

### LINKED ISSUE #156: WebIO does strange things and crashes when trying to render images
I've tried:

```julia
using TestImages, WebIO

testimage("cameraman") |> dom"div"
```

and the jupyter notebook more or less crashed while Juno survived and said that the result was a `Node` with `263144` descendants. Not sure what's going wrong but we definitely need to figure out image rendering.

rdeits:
This implements the suggestion I made in #156 (and fixes #156 ). 

It's kind of obnoxious to ask users to switch from `Node()` to `node()` in their code, so I'm open to other suggestions. 

For example, instead of doing this, we could keep the vararg `Node()` constructor as-is and replace all our internal usage with a new `_node_internal()` constructor that takes an `AbstractVector` of children. 

But in either case, I think we should be sure that the user-facing constructor has *only* the vararg method, rather than having the ambiguous vararg and Vector methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,10 @@ First, load the front end package (e.g. Blink or Mux; IJulia and Atom packages a
 Whenever a code cell returns a `WebIO.Node` object, IJulia will render it. For example,
 
 ```julia
-In[*]: Node(:div, "Hello, World")
+In[*]: node(:div, "Hello, World")
 ```
+
+The `node` (lowercase) function is a helper function which provides a convenient way to construct `WebIO.Node` objects.
 
 - On **Blink**
 
@@ -74,7 +76,7 @@ Return the WebIO Node from a web app to render it. Use `webio_serve` to serve th
 
 ```julia
 function myapp(req) # an "App" takes a request, returns the output
-    Node(:div, "Hello, World!")
+    node(:div, "Hello, World!")
 end
 
 webio_serve(page("/", req -> myapp(req)))
@@ -95,9 +97,9 @@ Let's say you want to display the following HTML:
 You can create a nested Node object:
 
 ```julia
-Node(:ul,
-    Node(:li, "get milk"),
-    Node(:li, "make a pie"), attributes=Dict(:class => "my-list"))
+node(:ul,
+    node(:li, "get milk"),
+    node(:li, "make a pie"), attributes=Dict(:class => "my-list"))
 ```
 
 `attributes` keyword argument sets the attributes of the HTML element.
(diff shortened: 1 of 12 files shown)


piever:
I'd actually prefer to have a `node` function. I've added one in Widgets already and I think this PR could extend that placeholder (this way people can use it without depending on WebIO when defining Interact recipes).
I would also like to be less strict about the tag and accept strings (i.e. `node("div", ...)`) if that doesn't create conflicts.

We also need to be careful what `n(v)` does when `n` is a `Node` and `v` a vector.

piever:
Regarding the switch to `node`, I think it's not too bad of in conjunction with the Julia 0.7 upgrade: People will be upgrading anyway and this one can be done with a simple Ctrl+ Shift+F or some at equivalent command on most editors I suppose.

shashi:
I like this change. It's a good solution for #156 although it would take 2 releases to fix. I think the deprecation you have added should serve OK in getting people to move to `node`. Better to do this now than later.

rdeits:
Ok, will do. I'm just adding a few more tests

rdeits:
@piever you mentioned having this `node()` function extend the one in Widgets.jl, but WebIO doesn't currently depend on Widgets. Do you want me to add that dependency?

piever:
Yes, that'd be ideal. The dependency is very lightweight and Widgets is already 0.7 compatible. Also, I hope the `node(s::AbstractStirng, args...; kwargs...) = node(Symbol(s), args...; kwargs...)` is not a problem to have in WebIO. In Widgets the macros by default replace symbols with subwidgets (like StatPlots, DataFramesMeta or JuliaDBMeta) so it's quite useful to be able to call this function with a string instead.

rdeits:
Ok, I've extended `Widgets.node()` and added a `::AbstractString` method to `promote_instanceof()`, so you can do `node(::AbstractString, children...)` and get the same result as `node(::Symbol, children...)`.

rdeits:
Ah, I just realized that extending `Widgets.node()` in this way might not be a good idea, as it's definitely type-piracy...

piever:
It is type-piracy, but it's the default strategy with low dependency packages. Plots does the same: the function `plot` is defined in RecipesBase and Plots extends it (and we do the same for `Observable`: the `show` method for `Observable` is defined here but WebIO doesn't "possess" neither `Observable` nor `show`). The only caveat is to be careful that WebIO is the package to define general methods for `node`.

rdeits:
ok

rdeits:
Just realized this is still open. Any objection to me merging it now?

piever:
I'm in favor of merging now and tagging so everybody can start updating.

PR #175, opened 2018-08-17
move widget methods from InteractBase

piever:
I'm moving methods from AbstractWidget here from InteractBase, so that it's easier to keep things in sync when the integration with other packages changes.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,4 +5,4 @@ Observables 0.0.3
 AssetRegistry
 Requires
 Compat 0.59
-Widgets 0.2.2
+Widgets 0.3.1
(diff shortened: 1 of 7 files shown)


piever:
This fails tests but I think it was failing already?

rdeits:
Yeah, the tests need the Blink and JSExpr PRs in order to pass. Sorry--it's hard to update when there are circular tests dependencies like this

piever:
I see, can I merge this one then? It's completely orthogonal, passes tests locally and I needed for the Interact update.

rdeits:
Yeah, that's fine

PR #180, opened 2018-08-20
fix undefined error in ijulia provider

piever:
Somehow this was overlooked when porting things to the new Requires package.


piever:
I still get `Could not open comm  -- Error: "Class webio_comm not found in registry "` in the jupyter console and widgets do not update, any idea what it could be?

rdeits:
I see this as well. I'm not sure what's going on, and I can't look into it right now. It does seem like either the `onjs` updates for the observables aren't being triggered, or they're being triggered but the IJulia connection messages aren't being sent.

PR #182, opened 2018-08-21
fix another missing IJulia prefix

rdeits:
@piever this is a PR into your branch from #180 . There was just another missing IJulia prefix (my fault) and the error message was lost since it was happening on some other task. With this, the basic ticking clock demo works for me in julia 0.7.


rdeits:
Actually the whole jupyter demo notebook is now working in Julia v0.7 with this PR.

piever:
You are on fire! Looks like everything's working then! This and loading css libraries were the last issues

PR #183, opened 2018-08-21
update travis

piever:


piever:
I'm temporarily checking out master on travis just to test that everything passes there, will add the proper lower bound in test/REQUIRE as soon as the Blink release is merged in metadata.

PR #185, opened 2018-08-24
Fix scope imports in iframes with Blink

### LINKED ISSUE #167: Scope imports (using AssetRegistry) don't work inside IFrames
For example, this works just fine:

```julia
julia> using AssetRegistry, WebIO, Blink

julia> s = Scope(imports=[AssetRegistry.register("/home/rdeits/.julia/v0.6/WebIO/assets/webio/dist/bundle.js")]);

julia> w = Window(); body!(w, s)
```

but this fails: 

```julia
julia> w = Window(); body!(w, iframe(s))
```

with a JS console error: 

```
40e9754…-bundle.js:2609 Uncaught (in promise) Error: Unable to resolve "/assetserver/40e9754662e3c699f1fcac299c40db93964f48fd-bundle.js" to about:blank/
  Resolving /assetserver/40e9754662e3c699f1fcac299c40db93964f48fd-bundle.js
  Loading /assetserver/40e9754662e3c699f1fcac299c40db93964f48fd-bundle.js
    at r (40e9754…-bundle.js:2609)
    at n (40e9754…-bundle.js:2609)
    at He.G (40e9754…-bundle.js:2609)
    at 40e9754…-bundle.js:2609
    at <anonymous>
```

This is the root cause of https://github.com/rdeits/MeshCat.jl/issues/49

cc @tkoolen

rdeits:
Fixes #167 
Fixes https://github.com/rdeits/MeshCat.jl/issues/49

The root cause turned out to be that the `iframe` document in Blink has its `baseURI` attribute set to `"about:blank"` which causes SystemJS to...I don't know. Do something stupid. 

I'm not sure, but I suspect this error came with the upgrade to electron 2.0, since I can't think of anything in our code that would have broken this. Firefox doesn't have the problem (it assigns the same baseURI to the iframe as the parent document). 

To resolve it, I'm just adding the `base` tag that matches the `base` of the containing document, which should force Blink/electron to do the same thing that Firefox does. 

I also removed the base URL detection logic from `iframe.jl`. I think that's now obsolete with the new `WebIO.baseurl[]`, correct?

### DIFF
--- a/assets/webio/dist/bundle.js
+++ b/assets/webio/dist/bundle.js
@@ -2864,7 +2864,7 @@ function doImports(scope, imp) {
         case "js":
             var cfg = {paths: {}}
             cfg.paths[imp.name] = imp.url
-            if (imp.url.slice(0, 4) === "pkg/" || imp.url.slice(0, 5) === "/pkg/") {
+            if (imp.url.slice(0, 12) === "assetserver/" || imp.url.slice(0, 13) === "/assetserver/") {
                 cfg.meta = {};
                 cfg.meta[imp.url] = {authorization: true};
             }
--- a/assets/webio/node.js
+++ b/assets/webio/node.js
@@ -249,7 +249,7 @@ function doImports(scope, imp) {
         case "js":
             var cfg = {paths: {}}
             cfg.paths[imp.name] = imp.url
-            if (imp.url.slice(0, 4) === "pkg/" || imp.url.slice(0, 5) === "/pkg/") {
+            if (imp.url.slice(0, 12) === "assetserver/" || imp.url.slice(0, 13) === "/assetserver/") {
                 cfg.meta = {};
                 cfg.meta[imp.url] = {authorization: true};
             }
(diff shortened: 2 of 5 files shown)


rdeits:
The `webcomponents-lite` loader was also broken in Iframes in IJulia. That should be fixed now as well.

hustf:
See comment in issue #167.

I'd like to assist if I can, but the first step is to check if things are basically running. I'm on Windows, and use both Google Chrome and Firefox, with this PR checked out.

1)
I was able to get simple examples working through Blink. There are some security messages regarding nodeJS integration, but it works just fine otherwise.

2)
Getting IJulia to work seems to require IJulia master. Even with master, the kernel locks up in some cases when using WebIO. I suspect there are Julia warning messages, but I don't really know how to retrieve them in a good way. No useful Julia messages are output by running 'jupyter notebook --debug' from the Anaconda prompt.

In a working notebook with Chrome, we get this ominous message which is of course unrelated to this PR:
```
jquery.min.js:2 [Deprecation] Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/
```

3)
Running Mux simple examples work just fine.

PR #187, opened 2018-08-27
WIP: fix Juno integration

pfitzseb:
This allows Juno to display WebIO objects but unfortunately uses Mux. It's also pretty slow... I'm pretty sure we don't want this merged in the state it's in, but hopefully someone can suggest improvements here :)

Upside is that it should work just fine in VSCode as well (with the correct mime type, obviously).

### DIFF
--- a/src/WebIO.jl
+++ b/src/WebIO.jl
@@ -56,7 +56,9 @@ function __init__()
         include(joinpath("providers", "mux.jl"))
     end
     @require Juno="e5e0dc1b-0480-54bc-9374-aad01c23163d" begin
-        include(joinpath("providers", "atom.jl"))
+        @require Mux="a975b10e-0019-58db-a62f-e48ff68538c9" begin
+            include(joinpath("providers", "atom.jl"))
+        end
     end
     @require Blink="ad839575-38b3-5650-b840-f874b8c74a25" begin
         include(joinpath("providers", "blink.jl"))
--- a/src/providers/atom.jl
+++ b/src/providers/atom.jl
@@ -2,6 +2,7 @@ using Logging
 
 const pages = Dict{String,Any}()
 const serving = Ref{Bool}(false)
+const port = Ref{Int}(9000)
 
 function routepages(req)
     return pages[req[:params][:id]]
@@ -21,29 +22,44 @@ function Base.show(io::IO, ::MIME"application/juno+plotpane", n::Union{Node, Sco
     id = rand(UInt128)
     pages[string(id)] = n
 
-    port = rand(8000:9000)
-
     if !serving[]
-        # hide http logging messages
-        with_logger(NullLogger()) do
-            @async begin
-                http = Mux.App(Mux.mux(
-                    Mux.defaults,
-                    Mux.route("/:id", routepages),
-                    Mux.notfound()
-                ))
+        setup_server()
+    end
+    print(io, "<meta http-equiv=\"refresh\" content=\"0; url=http://localhost:$(port[])/$(id)\"/>")
+end
 
-                websock = Mux.App(Mux.mux(
-                    Mux.wdefaults,
-                    Mux.route("/webio-socket", create_silent_socket),
-                    Mux.wclose,
-                    Mux.notfound(),
-                ))
+function setup_server()
+    port[] = rand(8000:9000)
 
-                Mux.serve(http, websock, port)
-                serving[] = true
+    # hide http logging messages
+    with_logger(NullLogger()) do
+        @async begin
+            http = Mux.App(Mux.mux(
+                Mux.defaults,
+                Mux.route("/:id", routepages),
+                Mux.notfound()
+            ))
+
+            websock = Mux.App(Mux.mux(
+                Mux.wdefaults,
+                Mux.route("/webio-socket", create_silent_socket),
+                Mux.wclose,
+                Mux.notfound(),
+            ))
+
+            for i = 1:100
+                iserr = false
+                try
+                    Mux.serve(http, websock, port[])
+                catch err
+                    iserr = true
+                    port[] = rand(8000:9000)
+                end
+                iserr || break
             end
+            serving[] = true
         end
     end
-    print(io, "<meta http-equiv=\"refresh\" content=\"0; url=http://localhost:$(port)/$(id)\"/>")
 end
+
+WebIO.setup_provider(::Val{:juno}) = setup_server()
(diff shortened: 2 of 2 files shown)


ajkeller34:
Thanks for your efforts to fix this. Can you elaborate on the vision for the recent changes to the Juno display system, or is that detailed somewhere? I was intending to write some code that would make use of the plot pane as an interface but I'm now wary to do so if there are going to be further changes made without carefully considered and tested deprecation paths.

pfitzseb:
I don't think Juno's plot pane display system will change anytime soon again. Most of the recent changes were prompted by https://github.com/JunoLab/Juno.jl/issues/91.

See [the Juno docs](http://docs.junolab.org/latest/man/info_developer.html#Displaying-Plots-and-Graphics-1) for information on how to use the new display system.

The missing deprecations (there wasn't any real deprecation path, so they would've just been errors) were a mistake, I agree, but I though I knew about most people who use Juno's Plot Pane and communicated the changes with them. Sorry for the inconvenience :)

src/WebIO.jl
```diff
@@ -56,7 +56,9 @@ function __init__()
         include(joinpath("providers", "mux.jl"))
     end
     @require Juno="e5e0dc1b-0480-54bc-9374-aad01c23163d" begin
-        include(joinpath("providers", "atom.jl"))
+        @require Mux="a975b10e-0019-58db-a62f-e48ff68538c9" begin
```
piever: Does this mean that the user has to manually type `using Mux` for this to work? Or is Mux somehow loaded by default in Juno?

pfitzseb: Yes, a manual `using Mux` is needed. We can't really load Mux here or in Juno by default. Ideally we wouldn't use Mux at all, but I'm not sure how that would look then.

piever: I wonder whether Interact could load Mux as well (taking the dependency) and become what used to be the old Escher package. Then the user would simply do `using Interact` and everything would already work.

cc: @shashi as I think he is planning to revive the old Escher as Interact+Mux instead

SimonDanisch:
What about this is slow?

pfitzseb:
Starting the webserver is slow.

SimonDanisch:
Ah, but you just do it one time right?

pfitzseb:
Sure. Still seems annoying though ;)

SimonDanisch:
Heh - btw, if I'm not mistaken, serve returns `nothing`....

src/WebIO.jl
shashi: So does Atom no more use Mux?

shashi: How about doing `using Mux` inside this require block?

shashi: Oh okay, of course. That's tricky.

shashi:
So is there a way to not use Mux? Does Atom integration have any way to hook into a server that can serve static files from disk?

pfitzseb:
So as far as I understand we only need to have some sort of socket for communication between Julia and the webview in the plot pane. Loading static files is trivial since we can run stuff in a full node context.

The last commit here tries to implement something like that, but unfortunately doesn't work since no connection is ever established. On the Atom side I'm getting
```
WebSocket connection to 'ws://localhost:8013/' failed: WebSocket opening handshake timed out
```
and Julia displays
```
e = ErrorException("Unexpected character\nLine: 0\nAround: ...GET / HTTP/1.1  Host: ...\n            ^\n")
```
I don't know about WebIO's internals to really figure out what's going on...

pfitzseb:
The (broken) Mux-less implementation now lives in https://github.com/pfitzseb/WebIO.jl/tree/sp/juno_alternative.

shashi:
Can we get Julia to install Mux in the build process of WebIO and Atom if we detect that the other package is installed? Would that be an OK thing to do in Pkg3's book?

pfitzseb:
I don't think so, at least not without either Atom or WebIO directly depending on Mux. 

AFAICT the minimal dependency to get things to work is WebSockets.jl, with HTTP.jl to enable AssetRegistry to work properly. And *some* routing would be beneficial as well, which is what Mux or Pages proivde...

piever:
I also have the feeling that Mux may not be the minimal set of dependencies for this to work, but it's very close. If we want other packages to have Juno (and in the future VSCode) integration via WebIO, maybe WebIO should just take the dependency? We could use the placeholder trick to move `render` to a low dependency package to allow external packages to overload it without heavy dependencies.

SimonDanisch:
I've a minimal pure HTTP + WebSockets implementation almost ready ;)

pfitzseb:
That still leaves the question about who should depend on those packages (either Mux or HTTP+WebSockets). Would you mind the Mux dependency here, @shashi?

shashi:
Yeah I think having the dependency here would not be ideal. I'm kind of concerned that people already think of webio as something too big to depend on just to define WebIO.render... we need to work in the reverse direction.

piever:
I also think WebIO is already a bit of a heavy dependency for some cases (it has a few dependencies some of which are not Julia).

To make a concrete example, I would like to support automatic LaTeX rendering for LaTeXStrings in Interact web apps. Defining the `render` method in Interact is type piracy / doesn't scale as a solution, but I also don't believe that LaTeXStrings should take a WebIO dependency.

I see two ways forward:

- We finish the placeholder strategy we were using with Widgets: `node` is already defined in Widgets, Interact widgets can be called from Widgets with `@nodeps slider(1:100)` for example and there is a `scope` function in Widgets. WebIO could extend it so that `scope(args...; kwargs...) = Scope(args...; kwargs...)`: if we also move `render` there (meaning, a placeholder), for most things it'd be enough to depend on Widgets. Packages that use WebIO by default to show their results (like PlotlyJS) would still take the dependency.

- We recommend Requires as a solution. FWIU Requires is a reasonably clean solution right now, so if a package simply wants to define `render`, they could do so in a `@require` block.

shashi:
I think for now I'm fine with requiring users to do `using Mux` in Atom. Is that ok @pfitzseb ?

pfitzseb:
Hm, I don't really like that much... The user shouldn't need to explicitly install a webserver just to get interactive plots etc.

I guess we might need to bite the bullet and have Atom.jl depend on WebIO and Mux (or HTTP and WebSockets if we go with @SimonDanisch's implementation).

pfitzseb:
I've migrated this code to Atom.jl now.

Issue #139, opened 2018-06-06
Cross-talk, warnings, and errors with multiple jupyterlab tabs


rdeits:
Jupyterlab allows multiple notebooks to be opened simultaneously in the same browser tab by using its own internal jupyter tabs. That seems to play very badly with WebIO.jl. For example, try running the `random_print_button()` demo from the Readme in two different jupyterlab notebooks (living in jupyterlab tabs in the same browser tab): 

```julia
function random_print_button()
    w = Scope()

    obs = Observable(w, "rand-value", 0.0)

    on(obs) do x
        println("JS sent $x")
    end

    w(
      dom"button"(
        "generate random",
        events=Dict("click"=>@js () -> $obs[] = Math.random()),
      ),
    )
end

random_print_button()
```

Running this code in both notebooks causes the first Julia kernel to receive the messages from *both* tabs, even though their scope IDs are different. The result is that the "JS sent $x" message is printed in the output of the first notebook even when you click buttons in the second notebook. 

Running the cell a few more times eventually gets WebIO into a state where neither button does anything except throw "invalid message received" warnings in the JS console and no output is printed at all.

PR #191, opened 2018-09-08
use a private RNG to generate scope IDs

rdeits:
this ensures that seeding the global RNG cannot result in scopes with
duplicate IDs


rdeits:
This also fixes some (formerly) mysterious duplicate scope ID warnings that used to be thrown when running the WebIO tests.

shashi:
Does this affect #139 at all (in a good way)?

shashi:
I guess it goes not.

PR #196, opened 2018-09-10
Setup IJulia communication from the JS side rather than Julia side

shashi:
This is now analogous to how all the other backends set up communication - from the browser side.

Should fix some flakiness which was being caused by bad ordering of events. Specifically errors such as:

```
No target "webio_comm" registered
WebIO.sendCallback not defined
```
cc @andreasnoack


rdeits:
Awesome. Do we need to do anything special to make this work inside iframes too?

shashi:
I just went over the iframe code, it should not make a difference.

Any idea what's up with CI?

shashi:
There is actually a job https://travis-ci.org/JuliaGizmos/WebIO.jl/jobs/426887317 but github is not showing anything under this PR.

mbauman:
Hurrah, I can confirm this fixed the issue I was seeing. Thanks!

PR #198, opened 2018-09-11
implement generic http provider

SimonDanisch:


src/providers/generic_http.jl
```diff
@@ -0,0 +1,147 @@
...
+function asseturl(file)
+    path = string(normpath(WebIO.assetpath), '/', normpath(file))
+    WebIO.baseurl[] * AssetRegistry.register(path)
+end
+
+function serve_assets(req, serve_page)
+    if req.target == "/"
+        return serve_page()
+    end
+    if haskey(AssetRegistry.registry, req.target)
```
pfitzseb: Maybe
```julia
function serve_assets(req, serve_page)
    response = serve_page(req)
    response === missing || return response
    if haskey(AssetRegistry.registry, req.target) 
        ...
```
instead so that the user of `WebIOServer` can implement routing themselves?

SimonDanisch: Yeah I was thinking about that ;) seems like a good change!

src/providers/generic_http.jl
```diff
@@ -0,0 +1,155 @@
...
+"""
+function WebIOServer(
+        default_response::Function = (req)-> missing;
+        baseurl::String = "127.0.0.1", http_port::Int = "8081",
+        verbose = false, singleton = true,
+        websocket_route = "/webio_websocket/",
+        logging_io = devnull,
+        server_kw_args...
+    )
+    if !singleton || !isassigned(singleton_instance)
```
shashi: is there any use to the `singleton` flag?

shashi:
Cool!

Can you add an example of this to the README?

It's not obvious to me how I can use this to serve multiple pages. Is there a way. The current Mux integration does support that.

pfitzseb:
> It's not obvious to me how I can use this to serve multiple pages. Is there a way? The current Mux integration does support that.

Yes, but you have to handle the routing yourself.

SimonDanisch:
Yeah routing is possible, although that's where it quickly starts to make sense to use Mux instead.
To be honest, this mainly has been an educational adventure for me, to figure out what Mux actually does under the hood ;) But now we also have an implementation with more control and less dependencies, so I thought it might be of value!

SimonDanisch:
taken over by https://github.com/JuliaGizmos/WebIO.jl/pull/204

PR #201, opened 2018-09-13
WIP: remove unsafe script

### LINKED ISSUE #200: Get rid of unsafe-script
This hack was ugly, annoying and fun but it's no more required. JupyterLab lets you execute JavaScript as long as the user says they trust it. See https://github.com/jupyterlab/jupyterlab/pull/4515

shashi:
I'm having the following problem here:

screenshot from 2018-09-13 17-51-29
@piever this situation arises with any simple `@manipulate` like

screenshot from 2018-09-13 17-56-57

```julia
@manipulate for i=1:10
   i
end
```

I'm didn't dig deep into why, but I feel like it probably shouldn't?

closes #200 when done


SimonDanisch:
unsafe-script seemed like it definitely makes problems

src/node.jl
```diff
@@ -104,12 +104,11 @@ function escapeHTML(i::String)
 end
 
 function Base.show(io::IO, m::MIME"text/html", x::Node)
-    write(io, "<div class='display:none'></div>" *
-          """<unsafe-script style='display:none'>
-          WebIO.mount(this.previousSibling,""")
-    # NOTE: do NOT add space between </div> and <unsafe-script>
-    write(io, escapeHTML(sprint(s->jsexpr(s, x))))
```
piever: Why don't we need `escapeHTML` any more?

piever:
It'd be great to get rid of `unsafe-script`. I'm not too sure I understand what's going on with this PR, but I think the initial behavior is problematic and seems like some issue with escaping?

src/node.jl
shashi: things in a `<script>` tag get special treatment.. for example, `&` is actually JS syntax but `&amp;` is not and will cause a parse error :upside_down_face: 

I think I figured out how to fix the issue here.

src/node.jl
```diff
@@ -35,8 +35,11 @@ because if they are printed inside a <script> tag, even if they are in quotes in
 a javascript string, the html parser will still read them as a closing script
 tag, and thus end the script content prematurely, causing untold woe.
 """
-encode_scripts(htmlstr::String) =
-    replace(htmlstr, "</script>" => "</_script>")
+function encode_scripts(htmlstr::String)
+    # sometimes a </script> can be inside a string inside a string inside a string...
+    # we need to go deeper....
+    replace(htmlstr, r"</(_*)script>" => x->replace(x, "/"=>"/_"))
```
shashi: @JobJob this hack was superb. But not ugly enough. :rofl:

shashi:
Ok my failing example works now, but this doesn't work for the initial output...

```julia
@manipulate for i=1:10
    i
    @manipulate for j=1:10
        i,j
    end
end
```
It might have to do with some code I edited in `WebIO.render` of Observable. But my edits look correct to me (there was some kind of double rendering going on before -- the nested observable value was both printed and serialized with the scope).

src/node.jl
JobJob: Lol, what's with the `_*`?

shashi: So it will match anything like `</script>`, `</_script>`, `</__script>` and add another underscore...

There's sometimes code that will call `setInnerHtml` which will in turn call `setInnerHtml` etc, so you need to escape just enough.

piever:
As a general comment, it's probably important to add extensive comments in the code explaining how the whole mechanism works, esp. wrt escaping and the various hacks, otherwise it's going to be an adventure the next time it needs to be updated!

PR #204, opened 2018-09-18
 implement generic http provider

SimonDanisch:
just the non forked version of #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,32 @@ end
 webio_serve(page("/", req -> myapp(req)))
 ```
 
+- Generic **HTTP**
+
+You can use the generic HTTP provider for any app - without the need to rely on WebIO.
+
+```julia
+
+# You can just create your own display function
+function Base.display(d::MyWebDisplay, m::MIME"application/webio", app)
+    println(d.io, "outer html")
+    # calling show will make sure a server is running and serves dependencies
+    # from AssetRegistry and a websocket connection gets established.
+    show(d.io, m, app) #<- prints the html + scripts webio needs to work into io
+    println(d.io, "close outer html")
+end
+# You can customize the server via the following environment variables:
+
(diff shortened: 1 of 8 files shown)


SimonDanisch:
I took the freedom to just depend on HTTP for now - mostly for easier testing.
If it's really against WebIO dependency policy, I can remove it.
But this now nicely decouples webio from any provider :) 
So WebIO doesn't need to depend on the provider and neither does the provider.
The provider can just define something like:
```Julia
    function Base.display(d::MyWebDisplay, m::MIME"application/webio", app)
        println(d.io, "outer html")
        show(d.io, m, app) # will make sure a server is running and serves deps + websocket etc
        println(d.io, "close outer html")
    end
```
All urls/ports etc are customizable by ENV variables:
```Julia
function global_server_config()
    if !isassigned(webio_server_config)
        setbaseurl!(get(ENV, "JULIA_WEBIO_BASEURL", ""))
        url = get(ENV, "WEBIO_SERVER_HOST_URL", "127.0.0.1")
        http_port = parse(Int, get(ENV, "WEBIO_HTTP_PORT", "8081"))
        ws_default = string(url, ":", http_port, "/webio_websocket/")
        ws_url = get(ENV, "WEBIO_WEBSOCKT_URL", ws_default)
        webio_server_config[] = (url = url, http_port = http_port, ws_url = ws_url)
    end
    webio_server_config[]
end
```
So any provider that can work with some html could get implemented with this.
If the provider also needs routing, we might want to figure out a way to register a callback that does the serving/routing part - shouldn't be a difficult addition ;)

SimonDanisch:
This test failure seems unrelated?!
```
Blink mocks: Error During Test at /home/travis/build/JuliaGizmos/WebIO.jl/test/blink-tests.jl:27
  Got exception outside of a @test
  IOError: connect: connection refused (ECONNREFUSED)
```

PR #205, opened 2018-09-25
readguarded not exported via require

SimonDanisch:


hustf:
Looking at the logs, it is clear that the failures occur because WebSockets isn't on the latest version of HTTP.jl, while you are also referring HTTP.jl methods directly. We'll remove that cap in, roughly, a week's time if not sooner.

The architecture of dependencies is of course not very nice when you run into this. I do not know the extent to which you use HTTP methods, but assume you have no special desire to change to the more failure-proof interface import WebSockets.HTTP.uzw. The upside to doing that would be fewer direct dependencies for this package.

The quick-fix here is adding a third argument to WebSockets.ServeWS: This argument specifies the logger directly: stderr. My favourite HTTP logger is, for the time being, Base.DevNullStream() and not stdout.

hustf:
When metadata is updated with WebSockets v1.0.3, try to rerun this?

SimonDanisch:
Thanks for the analysis ;) Maybe I should indeed start using `WebSockets.HTTP` for this... I'll give it a try!

hustf:
It would be interesting to see if this just runs flawlessly with no changes now.

SimonDanisch:
I'm not really sure why travis is still yellow - clicking on details shows all green:
https://travis-ci.org/JuliaGizmos/WebIO.jl/builds/433542699?utm_source=github_status&utm_medium=notification
I'm going to merge this!

PR #211, opened 2018-10-04
[WIP] Enormous rework of (mostly frontend) code.

### LINKED ISSUE #139: Cross-talk, warnings, and errors with multiple jupyterlab tabs

### LINKED ISSUE #203: Proposal: Consider Using Custom Jupyter Output Type
Currently, things go very wrong (in my experience) if a notebook that is using WebIO is refreshed. Ipywidgets (for example) behaves better because it's rendering system is more Jupyter-aware.

Here's a simplified version of (my understanding of) what ipywidgets does.
* You open a notebook with the ipywidgets nb-/labextension installed. The extension registers a renderer for the `application/vnd.jupyter.widget-view+json` MIME type.
* Say you run `import ipywidgets; s = ipywidgets.IntSlider()` (this displays a slider that can take integer values).
* The Python generates a display-data message with a `application/vnd.jupyter.widget-view+json` MIME type (which essentially just contains a reference to the widget's id) and sends a `comm_open` request to the browser associated with the id of that widget.
* The renderer mentioned above is called, looks up a global `widget_manager` object, tries to find the comm associated with the widget (if it can't, it displays an error) and subscribes to updates associated with the widget id assigned to the slider.
* **You refresh the page.**
* The extension initialization code does a check with the kernel (it sends a `comm_info_request`, which lists all the open comms) to see if any widget comms are registered.
* The notebook is initialized further, and the widgets are rendered via the same process described above.

We could (and I think should) do something similar (i.e. wrapping WebIO content in a `application/vnd.juliagizmos.webio.node+json`, or something similar, display MIME type; this would require a slight modification to IJulia itself, to allow displaying arbitrary MIME types, but nothing major).

This also has the advantage of not having to do all the `WebIO.mount` stuff inside the display data message, which seems kind of hacky. It would also not require the use of a custom asset registry (which I also think seems very hacky in the context of Jupyter; no other notebook functionality that I'm aware of hooks up its own asset server).

I'd be happy to help (I do consider myself to be _intimately_ familiar with the Jupyter protocol and a pretty decent web-oriented developer, much moreso that with Julia). I'm also happy to answer questions about why I think this is a good idea if you have any.

_I also confess that I have an ulterior motive in that I'm trying to develop a custom Jupyter web-based frontend (for an education-related project) and really want to be able to use Julia/Interact/WebIO, but find the current state of things untenable._

tag @shashi

travigd:
@shashi

I did a lot of stuff.

I tested this with Mux (and only implemented a Mux provider) and want to do more stuff with better integration with IJulia but I think this represents a better way of doing things as far as client side code is concerned.

Questions/comments/concerns appreciated. I'd really like to get something like this merged.

My own notes:
* I've injected some of my own frontend opinions; I think the biggest one was just not querying things by id since id should be globally unique in a document (I think I used a class name but we could also query on some data-* attribute like `node.querySelector('*[data-webio-scope="scope-id"]')`).
* I think the command used to update observables should be changed (I made a note of this in the message.ts file).
* I ended up playing around with the internal `WebIO.render` pathways of the Julia code and ended up introducing some stuff that I'm personally not fond of and would want to refactor (e.g. `renderScopePreferJSON`). The reason this is so ugly right now is because I was just having a hard time figuring out the complex ways in which render is recursively called (and I wanted to get rid of <script> tags that render things that then have <script> tags...)
* The escaping is now **much** simpler since all the data is passed as JSON. Escaping `</script>` simply becomes escaping all `/` characters (which is allowed since all `/` characters appear inside of strings and `\/` is interpereted as just `/`).

This will also require a very slight modification to JSExpr to not use a global WebIO instance. I modified JSExpr locally in two places and everything _just worked™_:

```diff
 function obs_get_expr(x)
     # empty [], special case to get value from an Observable
-    F(["WebIO.getval(", jsexpr(x), ")"])
+    F(["_webIOScope.getObservableValue(", jsexpr(x), ")"])
 end
 
 function obs_set_expr(x, val)
     # empty [], special case to get value from an Observable
-    F(["WebIO.setval(", jsexpr_joined([x, val]), ")"])
+    F(["_webIOScope.setObservableValue(", jsexpr_joined([x, val]), ")"])
 end
```
(`_webIOScope` is defined in the context where the function is evaluated).

When done, will close #203, #8. Supersedes #201.
JupyterLab plugin (which I'm more than happy to write/maintain) will close #139.

Again, I'm very open to comments and opinions and changing anything.


SimonDanisch:
Great work!
The Julia code changes seem pretty reasonable - would it be possible to put those in a different PR to make things a bit more incremental?
I have the suspicion, that no one here will be able to review this large diff thoroughly in a reasonable time frame - and I wouldn't want your PR to go stale!

travigd:
@SimonDanisch The Julia changes are relatively minor and aren't compatible with the previous version of the frontend code, unfortunately, so I don't think it makes much sense to separate them.

I'm also gonna fix up Blink then work on a JupyterLab extension. I'd like to also make a notebook extension which would increase stability of WebIO across page reloads but the current stuff I have implemented should (need to confirm!) work more-or-less as it does now (with the same level of flakiness-on-reload).

travigd:
Also pending [this IJulia pull request](https://github.com/JuliaLang/IJulia.jl/pull/755).

travigd:
I'm actually convinced that Blink isn't working as expected, at least when it comes to headless/containerized environments (which includes Travis).

I tried (using the official `julia:0.7` image with the current, unmodified versions of Blink and WebIO) to do the following:
```julia
Pkg.add("Blink")
using Blink
Blink.AtomShell.install()
w = Window(Dict(:show => false))
```
and got
```
ERROR: IOError: connect: connection refused (ECONNREFUSED)
Stacktrace:
 [1] try_yieldto(::typeof(Base.ensure_rescheduled), ::Base.RefValue{Task}) at ./event.jl:196
 [2] wait() at ./event.jl:255
 [3] wait(::Condition) at ./event.jl:46
 [4] stream_wait(::Sockets.TCPSocket, ::Condition) at ./stream.jl:47
 [5] wait_connected(::Sockets.TCPSocket) at ./stream.jl:263
 [6] connect(::Sockets.TCPSocket, ::Sockets.IPv4, ::Int64) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v0.7/Sockets/src/Sockets.jl:444
 [7] connect(::Sockets.IPv4, ::Int64) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v0.7/Sockets/src/Sockets.jl:428
 [8] #try_connect#5(::Float64, ::Int64, ::Function, ::Sockets.IPv4, ::Vararg{Any,N} where N) at /root/.julia/packages/Blink/vKGzM/src/AtomShell/process.jl:71
 [9] try_connect at /root/.julia/packages/Blink/vKGzM/src/AtomShell/process.jl:69 [inlined]
 [10] #init#6(::Bool, ::Function) at /root/.julia/packages/Blink/vKGzM/src/AtomShell/process.jl:85
 [11] #init at ./none:0 [inlined]
 [12] #shell#9(::Bool, ::Function) at /root/.julia/packages/Blink/vKGzM/src/AtomShell/process.jl:123
 [13] shell at /root/.julia/packages/Blink/vKGzM/src/AtomShell/process.jl:121 [inlined]
 [14] Window(::Dict{Symbol,Bool}) at /root/.julia/packages/Blink/vKGzM/src/AtomShell/window.jl:38
 [15] top-level scope at none:0
```

Probably related to [this Blink.jl issue](https://github.com/JunoLab/Blink.jl/issues/128).

travigd:
Ah, okay, I figured out all of my woes with Blink.jl (it's definitely not Blink's fault, no matter how much I wanted it to be...).

The issue is that it's passing on my machine™ because I have my own modified (as described above) version of JSExpr. It seems kind of weird that testing WebIO requires JSExpr considering how JSExpr depends on WebIO, so unless anyone has enormous objections, I'm going to remove the JSExpr testing dependency. This should also make it easier to update WebIO and JSExpr independently in the future.

piever:
Make sure to test Interact widgets before merging this: I see a lot of regressions. In Blink I can't show simple sliders:

```julia
julia> using Blink

julia> using Interact

julia> w = Window();

julia> body!(w, slider(1:100))
```
console error:
```
Uncaught (in promise) Error: not implemented
    at importResource (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:16878:13)
    at Array.map (<anonymous>)
    at _callee2$ (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:16855:75)
    at tryCatch (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:12751:17)
    at Generator.invoke [as _invoke] (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:12978:22)
    at Generator.prototype.(anonymous function) [as next] (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:12804:21)
    at http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:16718:67
    at new Promise (<anonymous>)
    at ../../webio/src/imports.ts.__awaiter (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:16695:10)
    at importAsyncBlock (http://127.0.0.1:6714/assetserver/c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:16847:10)
2c216e54911b0baacc0e30c0e7d3182f34c1d7bc8-blink.js:sourcemap:16451 Dispatching command "Basics.eval" not implemented.
```

whereas in Jupyter notebook the `slider` shows up (but not the value on the right), but `rangeslider` doesn't appear.

travigd:
@piever Whoops, my intention was not to suggest that this was really ready for testing (but it's getting closer and closer!).

My only remaining concern is... is rendering observables that contain DOM nodes a necessity? It's kind of hard to do this without a single WebIO global which is one of the reasons I made this PR. I see that Interact does this - I'm not sure if it's on purpose or not.

piever:
If you want to "replace a DOM node" dynamically in your app, the simplest solution is to put it in `Observable`. Say for example that I want to render `node1` if the user sets a checkbox to `true` and `node2` otherwise. In some cases I guess one can render all of them from the start and only set one of them to visible, but in general the `Observable{Node}` technique seems more powerful.

travigd:
Yeah, that's kind of what I thought (/was afraid) you were going to say.

So then I think the best attack strategy might be to have Observables be Nodes (in terms of the way that things are presented to the frontend code when `WebIO.render`'d).

So say you have this Julia code.
```julia
w = Scope()
counter = Observable(w, "counter", 0)
w(dom"div"(
    counter,
    dom"button"(
        "increment"
        ; events=Dict("click" => (@js function() $counter[] = $counter[] + 1 end))
    )
))
```

Currently, this is "transformed" into something like this node representation.
```
Scope (w)
  - dom"div"
      - Scope (contains counter, has js listener for when counter updates)
          - <pre>0</pre>
      - dom"button"
```

What happens then is that when an observable appears as a child of some node, it's set up with its own scope whose dom node is the html representation of that observable. This is tricky because then the HTML representation will be injected and will then try to use the WebIO global, which is bad news bears.

**So maybe we could have an Observable be rendered as a direct child of a node (without forcing it to be enclosed within its own scope).** This is the main point of what I'm trying to say in this long rambling comment.

The only issue I could foresee with this is that there would be issues if the observable rendered is not countained within the closest-ancestor scope. This is already an issue (on the current master of WebIO) if you try to use an observable within javascript while inside of a scope that doesn't contain that observable, and the simplest solution is to just attach all your observables to the same scope. However, what's not currently an issue (and what might become an issue) is that if you just render an observable as the child of a node, it's placed within it's own isolated scope that doesn't depend on whether-or-not that observable is within the ancestor scope.

So, question for @shashi (or anyone else who has thoughts/question):
* Does what I said above make sense (in the sense that you know what I'm talking about)?
* Is that an issue?

It seems like the "fix" would have to be something involving recursively finding observables which might not be that big of a deal...
```julia
function find_observables(node::Node)::Vector{Observable}
  observables = Vector{Observable}()
  for child in node.children
    if isa(child, Observable)
      push!(observables, child)
    elseif isa(child, Node)
      append!(observables, find_observables(child)
    end
  end
end
```
then attach all of the "found" observables to the scope when `WebIO.render`-ing it (though I'm not sure how that interacts with observable naming and whatnot).

So, assuming the above is not an issue or is taken care of at the very least, this is what I think would be the "optimal" solution.
```julia
w = Scope()
nodes = [dom"div"("hello"), dom"div"("goodbye")]
nodeobs = Observable(w, "node", rand(nodes))
w(dom"div"(
    dom"button"(
        "change the node"
        ; events=Dict("onclick" => (@js ...)) 
    ),
    nodeobs
)
```
would generate the following "node tree":
```
Scope (w)
- dom"div"
  - dom"button"
  - Observable
    - dom"div"("hello")
```
(whereas before, the `dom"div"("hello")` would have been rendered as a giant `WebIO.mount(...)` string before instead of just containing the JSON representation of the node).

## tldr
I hope the above made sense.

### Current Issue
When an observable that contains a node is rendered, that observable then contains the big `<script>WebIO.mount(...);</script>` thing which is not ideal (and won't work in Jupyter right now because it contains guards to not call WebIO.mount if it detects that the nbextension is installed).

### Proposed "Fix"
Modify the observable to contain a JSON representation of the node and don't wrap it in it's own scope.

### Addendum
We'd need to, I think, modify every rendered observable to become an observable of a node. For example,
```julia
w = Scope()
counter = Observable(w, "counter", 0)
WebIO.render(counter)
```
would then render a scope (`w`) that contains an observable child (`rendered_observable`) that contains a node (_e.g._ dom"pre"("0")). We'd set up a listener for counter that's like
```julia
map!(x -> render_as_node(x), rendered_observable, counter)
```

shashi:
```
 Scope (w)
  - dom"div"
      - Scope (contains counter, has js listener for when counter updates)
          - <pre>0</pre>
      - dom"button"
```
This is of course the easiest and most natural way to implement this. What we could do (I think this is what you suggest) is to make the underlying implementation morally better than `.innerHTML = "<script>WebIO.mount(...)</script>"`. Probably by calling `mount` on the object to render somehow directly.

>  if you try to use an observable within javascript while inside of a scope that doesn't contain that observable

I don't think this is the case... Every observable is identified by the (scope id, observable name) pair (see `@js $ob[]` for the JSON) no matter which observable it is embedded in. `this` in an observable's handlers will be the scope it is attached to...

> Addendum

We [already do this](https://github.com/JuliaGizmos/WebIO.jl/blob/7e74a492971e5f1ba76a472ef69991f98fc15b37/src/render.jl#L99) right?

packages/webio/src/DomNode.ts
```diff
@@ -0,0 +1,242 @@
...
+    namespace: DomNamespace;
+    tag: string;
+  }
+  props: DomNodeProps;
+}
+
+class WebIODomNode extends WebIONode {
+  readonly element: WebIODomElement;
+  children: Array<WebIONode | string>;
+  private eventListeners: {[eventType: string]: EventListenerOrEventListenerObject | undefined} = {};
```
shashi: Have you seen https://github.com/Microsoft/TypeScript/issues/21922 ?

shashi: Also any reason we need to keep these in the struct here?

travigd: I'm not sure what exactly you're referencing with the TypeScript issue. Do you just mean that we should use `EventListener` rather than `EventLIstenerOrEventListenerObject`?

And are you asking why we need to keep the event listeners in the object? I think it's so that we can `.removeEventListener` them if they change.

packages/webio/src/Node.ts
```diff
@@ -0,0 +1,95 @@
...
+export type WebIODomElement = HTMLElement | SVGElement | HTMLInputElement;
+
+// This was originally designed thinking that a Scope was a type of node,
+// but the thinking on that has changed.
+// And now, I'm thinking it has changed back.
+export type WebIONodeData = DomNodeData | ScopeNodeData | IFrameNodeData;
+export const enum WebIONodeType {
+  DOM = "DOM",
+  SCOPE = "Scope",
+  IFRAME = "IFrame",
```
shashi: Why do we need `IFrame` here? I think there was an example of a different package implementing its own `NodeType`. and extending `WebIO.NodeTypes[Foo] = {create:...}`, I don't know if it still does that, but hypothetically someone could want to do that. Why constrain this?

packages/webio/src/Node.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/**
+ * Abstract base interface for "node data."
+ *
+ * "Node data" is the data used to construct `WebIONode` instances and is the
+ * data that is passed over the comm from Julia to the browser.
+ */
+export interface WebIONodeDataBase {
```
shashi: This sounds like a "DataBase" :-p Why not call it "WebIONodeInterface" or "INode" or something like that

packages/webio/src/Node.ts
```diff
@@ -0,0 +1,95 @@
+import debug from "debug";
+const log = debug("WebIO:Node");
+
+import WebIO from "./WebIO";
+import WebIOScope, {ScopeNodeData} from "./Scope";
+import WebIODomNode, {DomNodeData} from "./DomNode";
```
shashi: This import seems backwards to me, shouldn't "Scope.ts" and "DomNode.ts" import "Node.ts" and kinda extend that?

Again, this would be possible if we didn't have the Unions below, I think?

packages/webio/src/Node.ts
travigd: Well the original reason was because iframes just have some different considerations than regular DOM nodes. In particular, I moved all of the iframe setup code out of `iframe.jl` and into the frontend.

I can modify things so that the construction of nodes is extensible though.

packages/webio/src/Node.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * An array of children (or null if not yet initialized).
+   */
+  abstract children: Array<WebIONode | string> | null;
+
+  /**
+   * The closest-ancestor scope associated with this node.
+   */
+  readonly scope?: WebIOScope;
```
shashi: Why do we need this?

packages/webio/src/Observable.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * Data that is sent from Julia/WebIO about the observable.
+ */
+export interface ObservableData<T = any> {
+  id: string;
+  value: T;
+  sync: boolean;
+}
```
shashi: Do we need an interface if there's only 1 type that implements it?

travigd:
So, regarding some of your comments, what I've written is based on the the idea that we shouldn't be pulling scopes out of thin air so that they behave more like scopes in the programming languages sense.

So basically, right now, every node maintains a reference to it's closest-ancestor scope and when it needs to look up or modify the value of an observable, it uses that scope to do so. I think I also should modify it so that it looks for observables in all ancestor scopes if it doesn't find it in the first ancestor scope.

So I **don't** think that this should work:
```julia
w1 = Scope()
w2 = Scope()
counter2 = Observable(w3, "counter2", 0)
w3 = Scope()
counter3 = Observable(w3, "counter3", 0)
w1(dom"div"(
  w2(counter2),
  w3(dom"button"("inc"; events=Dict("onclick" => (@js function() $counter2[] = $counter2[] + 1 end)))),
))
```
Currently, I (think?) it would work since all the scopes would exist on the page and all the scopes are located because they're registered at the top level.

Under what I'm proposing, it wouldn't because when trying to update `counter3`, we look for `counter3` observable in scope `w3` (and then, when we don't find it, in the parent scope `w1`). If that's a deal breaker, it's definitely possible to just lookup the right scope and then update its observable value but I think that kind of defies the notion of what a "scope" should be.

packages/webio/src/Scope.ts
```diff
@@ -0,0 +1,270 @@
...
+
+/**
+ * A map of observable names to arrays-of-listeners that should be
+ * evoked when the observable changes.
+ *
+ * `preDependencies` is a special array-of-listeners which are evoked
+ * when all the scope dependencies are loaded.
+ */
+export interface ScopeListeners {
+  preDependencies?: Array<(scope: WebIOScope) => void>;
```
shashi: This may not have any use anymore.

packages/webio/src/Node.ts
travigd: See comment below.

packages/webio/src/Scope.ts
```diff
@@ -0,0 +1,270 @@
...
+  [handlerName: string]: Array<((value: any, scope?: WebIOScope) => void)> | undefined;
+}
+
+export interface ScopePromises {
+  importsLoaded: Promise<any[] | null>;
+  connected: Promise<WebIOScope>;
+}
```
shashi: I had intended to make an extensible system for installing promises and listening to them...

Can we just use a super abstract type like a dict of promises?

packages/webio/src/Node.ts
travigd: Scope and DomNode do import and extend Node. I'll admit that it's kind of convoluted, but I think it's hard to do much better without losing the nice static typing. I'll try to play around with it.

packages/webio/src/Scope.ts
```diff
@@ -0,0 +1,270 @@
...
+     * }
+     *
+     */
+    handlers?: {
+      [observableName: string]: string[];
+    };
+
+    imports?: BlockImport;
+  }
+}
```
shashi: This is a reasonable use of interfaces :) I'm sorry if I seem like I'm hating on interfaces, but I believe they should not be used unless they actually reduce code substantially. It's clearer to read the whole struct definition in one place.

packages/webio/src/Scope.ts
```diff
@@ -0,0 +1,270 @@
...
+
+/**
+ * Handlers that are associated with the scope promises.
+ *
+ * @todo This needs to be refactored.
+ */
+export interface PromiseHandlers {
+  importsLoaded?: string; // (imports: any[] | null) => void;
+  // TODO: other promise handlers?
```
shashi: Yup.

packages/webio/src/Observable.ts
travigd: So this (`ObservableData`), `DomNodeData`, and `ScopeNodeData` exist only to describe the shape of the data that will be received from WebIO (via a call to `.mount` or on the comm). None of those are actually _implemented_ (in the sense of `class ... implements ...`) but I think they're useful nonetheless for static typing purposes.

packages/webio/src/Node.ts
```diff
@@ -0,0 +1,95 @@
...
+ * data that is passed over the comm from Julia to the browser.
+ */
+export interface WebIONodeDataBase {
+  type: "node";
+  nodeType: WebIONodeType;
+  children: Array<WebIONodeData | string>;
+}
+
+export interface WebIONodeParams {
+  scope?: WebIOScope;
```
shashi: I am not convinced it's a good idea to store this or that we need it...

shashi: At the least it will tax the GC right?

packages/webio/src/Scope.ts
travigd: Yeah, we could. So we could do something like
```ts
interface ScopePromises {
  importsLoaded: Promise<any[] | null>;
  connected: Promise<WebIOScope>;
  [promiseName: string]: Promise<any> | undefined;
}
```
which would keep the typing for those two promises.

packages/webio/src/IFrame.ts
```diff
@@ -0,0 +1,79 @@
...
+    iframe.src = `about:blank`;
+    iframe.frameBorder = "0";
+    iframe.scrolling = "no";
+    iframe.height = "100%";
+    iframe.width = "100%";
+    iframe.style.display = "block";
+
+    const {innerHTML} = iframeData.instanceArgs;
+    iframe.onload = () => this.initializeIFrame(innerHTML);
+  }
```
shashi: Ah, I see you moved this code to JS by defining a new node type.

(I remembered now that I was using the node type thing to create a React component). I somehow still feel like this feature is redundant with Scope itself, but haven't proved this to myself yet. But I guess we can keep it.

packages/webio/src/DomNode.ts
travigd: :woman_shrugging: I just copied that functionality from the current/master version of the frontend code. I changed the type to `EventListener` (will push soonish).

packages/webio/src/Observable.ts
shashi: It's kind of strange that they call both of the forms "interfaces" one use case is really more of a struct type?

In cases where an interface `X` is used as the type of an argument, does it actually compile to code which instantiates a class of type `new X(...)`?

travigd:
> A general comment about the "WebIO" prefix of exported call names, if they are all in the module's namespace do we really need the prefix?

I think a lot of it was to just avoid ambiguity with similar DOM-level concepts. For example, I think `WebIONode` is a better name than `Node`, and then `WebIOScope` and `WebIOIFrame` just followed for consistency.

travigd:
Also, re: scopes,

Do you think that my idea about hierarchical resolution of observables within scopes is good/is what we should do? Or should all scopes just be at the "top" level?

packages/webio/src/Node.ts
travigd: This is currently pending some design decisions about how hierarchical scopes should be, **but** no it won't have any real impact on the GC since every node only stores its parent scope.

I also need to add an event listener to delete the scope from WebIO if its element is deleted (but I think that the old WebIO just kept scopes around indefinitely and never deleted them until the page was refreshed -- so no matter what that'll be a performance gain in instances where scopes are created/removed frequently).

packages/webio/src/Observable.ts
travigd: Interfaces are completely abstract. Interfaces just describe the shape of an `Object`. So the only things that get "created" (in the sense of `new XXX`) are classes.

travigd: Which is to say - you can't construct an interface, but rather, you construct and object and label it as conforming to a certain interface.

```ts
interface MyInterface {
  myString: string;
  myNumber: number;
}
const myUnlabeledObject: MyInterface = {myString: "foo", myNumber: 42};
const myLabeldObject: MyInterface = {myString: "foo", myNumber: 42};
// Both will work in any places where TypeScript expects to see an object with interface MyInterface

**Comment by travigd** [comment]
I fixed some things. I renamed the data interfaces to schema to more accurately reflect what they're modelling, and also made a NODE_CLASSES object that is extensible.

**Comment by travigd** [comment]
Things are better™. I implemented an `ObservableNode` node type in Julia/TypeScript which is basically just a pointer to an observable contained within the scope.

@piever Interact works better... but I'm not sure what the current issue is.
I can do (in Blink because it's easiest to test interactively)
```julia
using WebIO, Blink, Interact
w = Window()
opentools(w)
@js w localStorage.debug = "*"
# Very simple toy example
node = (@manipulate for x in [1, 2, 3] x + 3 end)
body!(w, node)
```
and the actual nodes render... but they're not kept in sync. I think it's an issue with Knockout not being initialized correctly.

I checked out Knockout.jl in dev and added some `console.log`'s to the code... but I don't really know what's happening (oops). I'm very unfamiliar with Knockout. Maybe you could offer some pointers?

I tried manually updating one of the knockout observables
```js
WebIO.scopes["knockout-component-998b56dd-210b-45e1-a415-cf877ee542cf"].model.index(1)
```
and it triggers the updates (it sends it back to Julia which sends it back to the frontend with the updated value of `x + 3`) as expected. It looks like (and the "Event Listeners" tab of the chrome inspector confirms) the event listeners on the `<input />` are never set up... but again I don't know enough about Knockout to know how these things are **normally** done.

shashi:
`Node` is such a generic term that it's really confusing, we used to call `Reactive.Signal` `Node`, I think @SimonDanisch still uses that alias.

packages/webio/src/Observable.ts
shashi: But you can also `extend` an interface?

shashi:
> I would prefer leaving the behavior as-is.

The current behavior is that when an observable is created its identity is the same as its identity as a julia object (which we reflect by giving each observable a unique ID), the variable scoping rules of Julia hence apply to variables which point to the observable -- you can return them, use them elsewhere, set them elsewhere, just like you would a Ref.

piever:
> @piever Interact works better... but I'm not sure what the current issue is.
I can do (in Blink because it's easiest to test interactively)

If things aren't updating in Interact it may be simpler to start with Knockout. For example, does the following work:

```julia
incoming = Observable("")
on(println, incoming) # print to console on every update

template = node(:input, attributes = Dict("type"=>"text", "data-bind" => "value : message"))()
knockout(template, [:message=>incoming])
```

The way this works is that it creates a WebIO javascript observable that in turn gets synced to the Knockout version (see this code block: https://github.com/JuliaGizmos/Knockout.jl/blob/master/src/Knockout.jl#L44-L58). I would recommend adding some `console.log` statements in both syncing blocks (in the two directions) to figure what's happening.

travigd:
@piever I actually just figured it out :tada: 

The children nodes were being added to the parent node _after_ the `onimport` handler was being run, so knockout ran successfully but the DOM element it attached to was the empty div where the scope was mounted.

I'll push again in a few minutes (I'm just trying to get generic HTTP provider to work).

piever:
Do Interact things render correctly on this branch now? There have been a few complaints on Interact being unreliable in the notebook and I'm wondering whether I could point people to this branch for a more robust experience.

travigd:
@piever I believe so though I'd love some feedback. Ive only really been testing  toy examples. One thing to note might be the thing about JSExpr in the original merge request description.

travigd:
@piever Try now for notebooks. I've edited things so that WebIO will install the notebook extension as part of the build process.

piever:
Now in the notebook I get the widgets but they do not update the value and are not styled. Here's what I have in the console:

```julia
WebIO is registering SystemJS window global. main.js:17146:1
Loading extension: webio/main utils.js:37
Setting new WebIO window global. main.js:17343:3
Kernel: kernel_ready (7c1ed892-9aa0-4105-ada2-df917655e130) kernel.js:103
Setting new WebIO window global. main.js:17343:3
setTimeout executed 0 main.js:17443:3
Source map error: request failed with status 404
Resource URL: http://localhost:8889/nbextensions/webio/main.js?v=20181014215635
Source Map URL: main.js.map[Learn More]
```

travigd:
Things show up as styled and work for me in Blink (with that most recent commit after your comment just a moment ago).

If you want more (_a lot of_) debugging information from WebIO, you can do `localStorage.debug = "*"`. That'll persist across refreshes. You can do the same thing in Blink (though it won't persist) with
```julia
@js w localStorage.debug = "*"
```

piever:
Now it works much better! I especially like that it survives a page refresh and seems in general much more robust. 

I need https://github.com/JuliaGizmos/JSExpr.jl/pull/22/files for it to update successfully. Is there a less breaking way of doing this change? Can't `WebIO.getval` simply call `_webIOScope.getObservableValue` ? If we don't like the name we can deprecate it I guess, but removing it like this may cause updating issues (we can't update this without breaking JSExpr and viceversa).

One more thing: there used to be a hack that got removed. Things getting rendered into the notebook need to be inside a div of class `interactbulma`. Compare `node(:div, button(), className = "interactbulma")` with `button()` in this PR. The hack is not ideal but is the only way for interact to load invasive CSS like Bulma without screwing up the notebook (the fix is that on IJulia Interact loads a modified CSS that only acts inside divs of class `interactbulma` - see https://github.com/piever/BulmaThemes/blob/master/public/styles/wrap.scss). This should somehow be put back in (it used to be at https://github.com/JuliaGizmos/WebIO.jl/blob/17f88f69402ede3660df37d1e27995c0cebfdeec/src/node.jl#L117-L125).

As I assume the issue is quite general (for any library that wants to load some CSS in Jupyter) we could maybe come up with a better name for this and wrap all output to Jupyter in this class (without special casing `Widgets`). Maybe `jupyter-webio-output`?

shashi:
> we can't update this without breaking JSExpr and viceversa

Well JSExpr still depends on WebIO so we can upperbound the dependency in all previous versions, and release a new version with a lower bound.

shashi:
Good job on getting tests passing ! :1st_place_medal: :tada:

travigd:
@piever I can shim `WebIO.getval` for now (I'll also add a `console.warn` statement about deprecation). And ah, I see, I didn't implement the hack for rendering a Widget using the `WEBIO_NODE_MIME` MIME type. I can do that.

We can also definitely just wrap everything rendered by WebIO in a `.webio-mountpoint` class (I don't think this would be dependent on just Jupyter, and if you wanted to specifically select just Jupyter, you could do `.notebook_app .webio-mountpoint`). That sound good? Or like `.webio-rendered`. I'm not attached to the classname itself.

I'm starting to see the light at the end of the tunnel. :joy: :tada:

My current todos:

- [ ] Switch back from hierarchical scope resolution to global (relative to the WebIO instance) scope resolution (after conversation with @shashi)
- [ ] `interactbulma` hack
- [ ] Internal naming fix-ups (e.g. complete migration to `Schema`)
- [ ] Complete `imports.ts` (and maybe try to make it less mired in details).
- [ ] ???
- [ ] Profit

In the grand scheme of things™, I'll implement a JupyterLab plugin. That can be done after this pull request though is merged, though. I think.

piever:
I'd like to get rid of the interactbulma hack. If it's easy enough to add the `webio-mountpoint` class to all things rendered by WebIO I propose the following:

1. add the class `webio-mountpoint interactbulma` to the top level div added by WebIO
2. I'll adjust InteractBulma to work with the `webio_mountpoint` instead
3.  remove the class `interactbulma` after a while

travigd:
After that push (`bbda73e`: "Use global scope resolution of observables."), I believe WebIO should work with the older (aka not-modified-by-me) version of JSExpr as well.

piever:
A couple of feedback from the Interact side:

1. this somehow breaks `tabulator` from InteractBase, say:
```julia
 tabulator(OrderedDict("a" => 1, "b" => 2, index = 1)
```

Now when clicking on different buttons the value on screen (1 or 2) doesn't get updated.

2. This is something that has bugged me for a while (not a regression introduced here, but somehow I'm hoping that this PR could fis this). If this is totally unrelated, sorry for the noise!

There are some strange issues when rerendering widgets. If I start with:

```julia
using Blink
w = Window()
v = Observable(20)
s = slider(1:100; value = v)
body!(w, s)
```
the slider marks `20`. If I move to another number, the `s` object gets modified, but on rerendering:
```julia
body!(w, s)
```
the displayed value goes back to `20` (even though `s[]` correctly gives the updated number).

shashi:
>  If I move to another number, the s object gets modified

Strange. Do you mean you're calling `body!(w, s)` after setting `s[] = 25`?

What happens if you do `observe(s)[] = 25` and then render?

piever:
> >    If I move to another number, the s object gets modified
> 
> Strange. Do you mean you're calling body!(w, s) after setting s[] = 25?
> 
> What happens if you do observe(s)[] = 25 and then render?

Same effect, it still re-renders to the original value: no idea how that's possible...

travigd:
I think that's a Julia side fix. The _setup_scope command should cause Julia to send the current value of all the observables within that scope to the client.

travigd:
@piever 
The issue is that somehow, something within the tabulator widget is getting `WebIO.rendered`'d before it gets to the final before-output stage, so it contains a nested `<script>WebIO.render(...);</script>` which doesn't work since we let Jupyter invoke `WebIO.render` (that way it can use the correct instance of `WebIO`).

travigd:
I looked into it a little bit more and it seems(?) that it's happening because you have an observable that contains a scope. I can continue looking into it a bit later today I think.

piever:
>  it seems(?) that it's happening because you have an observable that contains a scope.

It's kind of crucial that this works though, because every Interact widget is its own scope and it's quite common to put them inside `Observable` to be able to update the structure of the UI dynamically.

travigd:
@piever Try now.

There were two factors that were inhibiting things from working.

* Scopes within observables weren't being rendered (first commit, `Render scopes within observables.`)
* Observables weren't being locally synced -- what I refer to in the code I wrote as "reconciliation" (second commit, `Implement observable reconciliation.`)

The brilliant and wonderful @shashi pointed the latter out to me a while ago and, as the devoted contributor I strive to be, I promptly forgot.

piever:
Thanks! That fixes the example above, I think more complex cases are still broken, will try to come up with a MWE soon.

travigd:
Current status:
- [ ] Fix `display(...)` in Jupyter (this is an issue with the extension code and probably something to do with internal differences between the notebook's handling of `display_data` and `execute_result`)
- [ ] Copy #217

piever:
I still see some issues with Interact in more complex scenarios. This is not quite a MWE as it requires the unregistered package [TableWidgets](https://github.com/piever/TableWidgets.jl).

```julia
julia> using Blink, Interact, TableWidgets, DataFrames

julia> df = DataFrame(x = 1:10);

julia> w = Window()

julia> wdg1 = addfilter(df);

julia> wdg2 = dataeditor(df);

julia> ui = tabulator(OrderedDict("a" => wdg1, "b" => wdg2));

julia> body!(w, ui)
```

screenshot from 2018-11-02 18-42-28

Basically here I'm doing a tabulator whose items are each pretty complex (each has its own scope and may be composed of many subwidgets) and somehow it seems that both are displayed one on top of another instead (regardless of which "tab" of the tabulator is selected). If that can be of any help, `tabulator` works by rendering all items but only setting one to visible. Sorry for only having a complex example, but it seems like simple stuff is already working quite robustly, which is definitely good news :)

piever:
As a double check, I would also try the more complex (depending on [TableWidgets](https://github.com/piever/TableWidgets.jl):

```julia
julia> using Blink, Interact, TableWidgets, DataFrames

julia> df = DataFrame(x = 1:10);

julia> w = Window()

julia> wdg1 = addfilter(df);

julia> wdg2 = dataeditor(df);

julia> ui = tabulator(OrderedDict("a" => wdg1, "b" => wdg2));

julia> body!(w, ui)
```

before merging as it seems to fail in a slightly different way than the buttons (here both options are always displayed one on top of the other - with the buttons instead if I click on "b" the first button disappears as it should).

travigd:
@piever I think it works for me now. The issue was BIZARRE; it's not that important but...

I was looking at the `importsLoaded` handler (which executes all the Knockout setup code) and my code (falsely) assumed it was a string (when it's actually a list of strings). So it did
```js
// Code operated under assumption that this was a string, not array
const importsLoadedSource = ["function() { ... }"];
importsLoaded = eval(`(${importsLoadedSource})`);
importsLoaded(dependencies);
```
which caused all kinds of weird nondeterminism that I admit I do not understand.

piever:
Great job! It seems like even pretty complex GUIs work reliably now (actually, more reliably than they used to...). Maybe this can be merged soonish and the Jupyter plugin happen in a separate PR? The diff already looks massive...

shashi:
:tada: :clap: :cake: :beers:

PR #231, opened 2018-11-20
Fix SVG support.

### LINKED ISSUE #226: svg syntax broke with new update
The following used to work to produce an svg:

```julia
n = 10
color = "yellow"
h, w = 100, 100
attributes = Dict(
    "fill" => color,
    "points" => join(["$(w/2*(1+sin(θ))),$(h/2*(1+cos(θ)))" for θ in 0:2π/n:2π], ' ')
)
dom"svg:svg[width=$w, height=$h]"(dom"svg:polygon"(attributes=attributes))
```

but it no longer seems to work (I'm on WebIO 0.6.0).

It was originally reported on Interact at https://github.com/JuliaGizmos/Interact.jl/issues/278 (curiously I do not see the same error but still cannot see the svg output).

travigd:
Fixes #226.


shashi:
OK to squash and merge :+1:

PR #232, opened 2018-11-22
Remove Blink @requires thing

ranjanan:
I don't know which should be merged first, this or https://github.com/JunoLab/Blink.jl/pull/174. But we could merge both and tag new releases.


travigd:
I would think (please correct me if I'm wrong!) [JunoLab/Blink.jl#174](https://github.com/JunoLab/Blink.jl/pull/174) should be merged and released first since that change isn't breaking where this change is (that is, upgrading only Blink doesn't upgrade Blink+WebIO but only upgrading WebIO would break Blink+WebIO).

Is there any way to check the installed version of a package that we don't depend on (that is, a package that isn't listed in `REQUIRE` without Julia complaining about it)? If so, we should definitely issue a warning about upgrading the version of Blink.

NHDaly:
> Is there any way to check the installed version of a package that we don't depend on (that is, a package that isn't listed in `REQUIRE` without Julia complaining about it)?

`Pkg.installed()["Blink"]` seems to work even if it's not in the `REQUIRES`. As long as it's installed, that seems to work. Notably, it _doesn't_ work when `Pkg.test`ing a package (because it copies a new Manifest file for running the tests, and so in that project Blink isn't installed). So perhaps something like this would work?:

```julia
using Pkg
pkgs = Pkg.installed()
blink_version = get(pkgs, "Blink", nothing)
if blink_version != nothing && blink_version < v"XX.XX.XX"
    error("...")
end
```

travigd:
Closing as stale.

This is definitely still worth addressing but should be evaluated against v0.8 (which should be released very soon).

PR #235, opened 2018-11-29
Fix @webio/webio package.json.

travigd:
I need to use WebIO as a dependency in my application (it's a specialized Jupyter client). These changes allow that (notably by adding the correct values for main and types in package.json).

We may also consider, in the future, publishing these to NPM under the @webio organization.


travigd:
I'm not sure why code coverage changed... I didn't edit any `.jl` files.

piever:
I wouldn't worry about it: I think coverage used to be extremely generous in deciding which lines were tested and is now much stricter, most likely this has nothing to do with your PR.

travigd:
@shashi and I were also talking about the fact that the diff for this (pretty tiny) commit is ENORMOUS on account of the generated JavaScript.

My proposed solution was to just minify so that GitHub just reports a one line change (and the CLI will report a "binary file changed" on account of the gitattributes). Does that seem okay?

travigd:
Addressed in `webio-retool` branch.

Issue #234, opened 2018-11-29
Use Comm to Communicate Dependencies on IJulia 


travigd:
Imports should be communicated via comm on IJulia. This is useful for situations where the kernel and the web service don't exist on the same machine (e.g. using [Jupyter enterprise gateway](https://github.com/jupyter/enterprise_gateway)) and also removes the need to install a Jupyter notebook (python) plugin.

shashi:
I would advise against this actually because the browser cannot load assets asynchronously in this setup.

travigd:
How so?

travigd:
I think the throughput will be bounded by the network right? And we can bundle the files in the comm request which is a good thing (e.g. webpacking is more effecient than multiple files).

Only issue is binary files won't be transferred efficiently since they'd have to be b64 encoded.

Also I do think it's generally a good idea™ since the kernel and notebook aren't always on the same server.

Issue #216, opened 2018-10-22
Remote procedure calls in WebIO


piever:
There was some conversation on slack on implementing generic RPC's in WebIO (meaning, from javascript to call Julia, wait for the result  and then do something with it). I'm copying it here as the idea is interesting and things always risk getting lost on slack.

The proposed syntaxes by @travigd is:

```julia
myrpc = WebIO.RPC(myfunction)
```
 and then 
```julia
js"async function (x) { let result = await $rpc(x); /* other stuff here */ }"
```

travigd:
Maybe we can also generalize this to "local procedure calls" (so to speak).

```julia
addFunc = WebIO.Func(js"""
function add(x, y) {
  return x + y;
}
"""
```
and then
```julia
js"myVal = await $addFunc(3, 4);
```

Keeping it await-based would avoid some issues about when to transfer function bodies to the JS side (e.g. when do we send the body of `addFunc` to the browser?). Only downside is that async isn't globally supported but IS [supported by all recent browsers](https://caniuse.com/#feat=async-functions). We could also (in the _distant_ future) add a babel transpilation plugin.

This could also make for an extensible way for other libraries to provide WebIO functionality.

shashi:
I'd just like to add that we can simply use Conditions to have the `jseval` function that executes some JS in a Scope and waits till a result is returned (if the result is a Promise then waits on the promise resolution and returns the promised value). It would look something like:

```julia
jseval(scope, @js ...)
```
Of course this is just the Julia side of the API.

shashi:
Just remembered we already have `evaljs` which does not wait, but we can I guess add `evaljs_sync`. Also this has been there in Blink for a long time `@js` -- sync and `@js_` -- async.

shashi:
```js"async function (x) { let result = await $rpc(x); /* other stuff here */ }"

PR #238, opened 2018-12-23
Sane-ify messaging protocol and start implementing request-response messaging.

travigd:
I'm trying to make some of the messaging protocol more consistent and sane, as well as implement request-response style messaging (this is necessary for #216 and #234).

## Issues
- [ ] `await_request` hangs the kernel if you try to `fetch` or `take!` the channel associated with it

@shashi @rdeits If you have some time I'd appreciate your help (both with a bit of review and for ideas on the bug) on this.


src/scope.jl
```diff
@@ -259,24 +260,61 @@ function render(s::Scope)
...
+    request_id = string(rand(UInt64))
+    message = Dict(
+      "type" => "request",
+      "request" => request,
+      "requestId" => request_id,
+      "scope" => scope.id,
+      data...
+    )
+    send(scope.pool, message)
+    return await_response(request_id)
```
shashi: between the previous line and this there can be a situation where the message arrives and is discarded because there's no await right?

The right way to do this kind of one-time wait in Julia is using Condition. Channels actually use this to serve a different purpose.

shashi:
wait so how did you solve the event loop getting stuck issue?

travigd:
@shashi I didn't, but it's not really a regression from the way things worked before. Previously, evaljs didn't support returning a value (and if you use it without expecting a value, things work exactly the same way as before).

In IJulia/Jupyter,
```julia
# Cell 1
s = Scope()
s(dom"div"("hello"))

# This works (cell 2)
evaljs(s, """console.log("Hello!")""")

# This works (cell 3)
@async begin
    url = fetch(evaljs(s, """window.location.href"""))
    println("Current url is $url")
end

# This doesn't work in IJulia, but does work in other providers (tested with Blink).
# In IJulia, **this bricks the event loop.**
url = fetch(evaljs(s, """window.location.href"""))
println("Current url is $url")
```

shashi:
Ok I guess we can live with that! :)

Issue #229, opened 2018-11-20
Dependencies Rework (RFC/Proposal)


travigd:
This originated in the Gizmos slack. We were talking about altering (or even getting rid of) "scopes" since they don't seem to be a rigidly defined concept. We could replace everything with DOM and Observable nodes. This would require reworking dependencies/imports though.

## Proposal
I proposed something like the following, together with an `onmount` function that works for any node.
```julia
my_div = dom"div"("Loading...")
mydep = WebIO.Dependency("https://.../my-cool-library.js")
onmount(my_div,
    js"""
    async function() {
        const mydep = await $mydep;
        mydep.someFunction();
    }
    """)
```

I also suggested having support for automatically downloading NPM modules. This wouldn't require NPM itself since (I believe?) the modules are downloaded as tarballs. This could be (should be?) made into another package.

```julia
using WebIONPM
React, ReactDOM = NPMDependency.(["react@^16.6.0", "react-dom@next"])
myjs = js"""
async function() {
    const [React, ReactDOM] = await Promise.all([$React, $ReactDOM]);
    // Etc etc etc
}
"""
```

## JSExpr.jl
Finally, I'm not sure how this would integrate into JSExpr. We could steal the await syntax again I suppose (just generally integrate `@await` into JSExpr for any asynchronous function).

```julia
onmount(my_div) do
    @var React = @await $React
    @var ReactDOM = @await $ReactDOM
    # ...
end
```

Another syntax might by using a RequireJS style API.
```julia
onmount(my_div, @require [React, ReactDOM] do React, ReactDOM
    ReactDOM.mount(...)
end)
```
But this seems noisy.

## Async Support
`await` isn't supported by every browser, but [it supported by all of the latest browsers](https://caniuse.com/#feat=async-functions) (>85% overall). Given that we're targeting a more technical audience, I think this might be okay.

## Async Ordering
Asynchronous code can be harder to manage and cause race conditions, especially with `onmount` and parent/child relationships. Currently, _I believe that_ child `onimport` calls are completed before the parent `onimport`.
We could either say "good luck, you're on your own" with regards to this ordering or we could guarantee some execution ordering (e.g. we wait for children `onmount`'s to finish before triggering the parent `onmount`). We could also do an `await this.promises.childrenMounted` type thing.
I think the best thing to do would be to have an `onmount` function that takes a `wait_for_children` type argument (which defaults to true).
```julia
# These would be equivalent
onimport(my_div, js"""
[...]

Issue #225, opened 2018-11-15
IFrame scope issues in 0.6.0


rdeits:
Hi all! I'm having some issues with MeshCat.jl using WebIO 0.6.0, and I think the problem comes down to the new iframe behavior. As an example, try running the p5.js demo in IJulia, using an iframe to wrap its output: 

```julia
using WebIO, JSExpr

w = Scope(imports=["//cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/p5.js"])

onimport(w, @js function (p5)
    function sketch(s)
        s.setup = () -> s.createCanvas(640, 480)

        s.draw = function ()
          if s.mouseIsPressed
            s.fill(0)
          else
            s.fill(255)
          end
          s.ellipse(s.mouseX, s.mouseY, 80, 80)
        end
    end
    @new p5(sketch, this.dom.querySelector("#container"))
end)

iframe(w(dom"div#container"()))
```

I see no output at all from this when using WebIO 0.6.0. From poking around at the similar issues with MeshCat.jl, I think the problem is that previously we were running JS code inside the iframe's scope (with `window` and `document` set to those of the iframe), but that no longer seems to be the case.

travigd:
I've been looking at this for... a while.

I think the issue boils down to this [line in p5](https://github.com/processing/p5.js/blob/edb41cf513efafd3734221a3b8cdc266f90d6bb3/src/core/main.js#L342):

```js
// this is the p5 instance
this._setup = function() {
    // ...
    var canvases = document.getElementsByTagName('canvas');
    for (var i = 0; i < canvases.length; i++) {
      // unhide canvases
    }
}
```

P5 hides canvases when they're created (to avoid a "flicker" while they're being setup) and then unhides them after the setup is done.

However, (**and I'm about 90% sure of this**), document in this context refers to the document of the parent window and not the document of the IFrame. The reason for this is because p5 is loaded in the top window context (since that's where WebIO is loaded and that's where SystemJS "lives").

To be more explicit, in a Jupyter notebook,
* We render some WebIO output in the notebook
* The nbextension activates and creates a new `WebIO` instance which is created in the notebook's Window
* The WebIO rendering code creates an IFrame and _patches in_ the singleton WebIO instance (_i.e._ `const iframe = ...; iframe.contentWindow.WebIO = window.WebIO;`)
* A scope is created in the IFrame.
* The scope requests P5.
[...]

PR #242, opened 2019-02-04
[WIP] WebIO retool branch

shashi:
This is a very large pull request (oops). It shouldn't immediately break anything that exists today (possibly with the exception of IFrames, because that's a mess, à la #225, and seems to be hard to fix in a reasonable way). Notable, Interact.jl should still work (with some deprecations).

## Major Goals
- [ ] Dependency rework (`WebAsset` model, #229)
  - [x] `JSAsset`
  - [x] `CSSAsset`
  - [ ] `HTMLAsset`
- [x] Clean up comm messaging API
  - [x] Request / Response Messaging
- [x] Jupyter Lab Provider


src/imports.jl
```diff
@@ -91,8 +91,8 @@ lowerdeps(xs::AbstractArray) = Dict(
 lowerdeps(x::Sync) = Dict("type"=>"sync_block",
                           "data" => map(lowerdeps, x.xs))
 
-function import!(scope, x)
-    push!(scope.imports, x)
-end
+# function import!(scope, x)
+#     push!(scope.imports, x)
+# end
```
shashi: Is this a temporary thing?

shashi: ah I see the new `onmount` now, all good!

travigd:
## To-Do's
- [ ] Add HTML imports

piever:
I'd like to have an InteractBase branch to update InteractBase to the new WebIO. The main thing missing seems to be css import: what is the correct way to import css or js in a scope here?

shashi:
forgive my dumb question: `tsc -p .` fails with 
```
jupyter-lab-provider/src/labextension.ts(1,19): error TS2307: Cannot find module 'debug'.
mux-provider/src/mux.ts(1,19): error TS2307: Cannot find module 'debug'.
webio/src/DomNode.ts(1,26): error TS2307: Cannot find module 'debug'.
webio/src/IFrame.ts(1,26): error TS2307: Cannot find module 'debug'.
webio/src/Observable.ts(1,26): error TS2307: Cannot find module 'debug'.
webio/src/ObservableNode.ts(1,26): error TS2307: Cannot find module 'debug'.
webio/src/Scope.ts(2,26): error TS2307: Cannot find module 'debug'.
webio/src/WebIO.ts(3,19): error TS2307: Cannot find module 'debug'.
webio/src/events.ts(1,19): error TS2307: Cannot find module 'debug'.
webio/src/imports.ts(2,26): error TS2307: Cannot find module 'debug'.
```

what should I do?

travigd:
@piever That's not a public API deprecation and it should be fixed within WebIO. The issue was that messages looked like `{"command": "observable-name", ...}` to update an observable with other special-case commands like `{"command": "_setup_scope", ...}` and I shifted it to be more consistent; so instead of the command being the observable name (_implicit_ observable update), it's now `{"command": "update-observable", "name": ..., "value": ...}`. So the WebIO client side code is still sending the old style message.

Maybe it's possible that the nbextension didn't get updated? It definitely shouldn't occur normally.

src/webasset.jl
```diff
@@ -0,0 +1,56 @@
...
+    extension = lowercase(split(name, '.')[end])
+    if extension == "css"
+        return CSSAsset(name)
+    elseif extension == "js"
+        return JSAsset(name)
+    end
+    error("Unknown import extension: $(extension).")
+end
+
+jsexpr(io, assets::Vector{WebAsset}) = jsexpr(io, map(jsexpr, assets))
```
shashi: `jsexpr` is not what we should use to generate JSON here though. It should be `JSON.lower`, see imports.jl. I think this file shares too much with imports.jl that they should be 1 file...

src/scope.jl
```diff
@@ -116,13 +118,23 @@ function Scope(id::String=newid("scope");
         private_obs::Set{String}=Set{String}(),
         dependencies=nothing,
         systemjs_options=nothing,
-        imports=[],
-        jshandlers::Dict=Dict()
+        imports=nothing,
+        jshandlers::Dict=Dict(),
+        mount_callbacks::Vector{JSString}=Vector{JSString}(),
```
shashi: OK, so there should have either been a default callback here that loads the import (like the one in `onimport`)... Or we should handle this on the JS side, because sending this callback for every scope can be avoided.

Right now, importBlock is only called for `scope.instanceArgs.imports` which you have commented out below. Uncommenting that and implementing JSON.lower should fix the issue. Then `onmount` should be a purely renaming deprecation.

This is the reason why imports are not _reaally_ working right now: they only work if there is an `onimport` callback on the scope -- but we want it to work even otherwise.

shashi:
were there some functional goals you were going for when you introduced JSAsset and CSSAsset types? or was it just trying to give them types?

travigd:
It was to enable the `await $myasset` syntax that allows you to use imports in (for example) callbacks.

shashi:
Oh ok!

travigd:
Haven't seen that green checkmark in a long time. :smile_cat: 

I'm going to merge this and we can continue to work on thing incrementally instead of all being contained within this PR.

shashi:
Okay I'll reiterate what I said on slack. Let's make small PRs implementing a single small feature from now on. We need to make reviewing easier!!

PR #250, opened 2019-03-28
Fix use of removed jsexpr in IJulia.

travigd:


src/providers/ijulia.jl
```diff
@@ -44,13 +45,12 @@ function main()
     script_id = "webio-setup-$(rand(UInt64))"
     warning_div_id = "webio-warning-$(rand(UInt64))"
     warning_text = "Loading WebIO Jupyter extension on an ad-hoc basis. Consider enabling the WebIO nbextension for a stabler experience (this should happen automatically when building WebIO)."
-    display(HTML("""
```
shashi: Do we still need this stuff? I think we can expect users to be willing to run `build` (we can just print that), Pkg.update automatically runs build anyway

shashi: ??

PR #253, opened 2019-03-28
Fix @js_str interpolation issues.

### LINKED ISSUE #111: js"..." does not escape dollars correctly
```julia
julia> js"this.\$refs"
WebIO.JSString("this.\\\$refs")
```

whereas the correct thing is I believe `WebIO.JSString("this.\$refs")`.

travigd:
Fixes #111.

Requesting review from @shashi and @piever because writing idiomatic Julia isn't exactly my strong suit.

### DIFF
--- a/src/syntax.jl
+++ b/src/syntax.jl
@@ -56,6 +56,24 @@ end
 Base.print(io::IO, x::JSString) = print(io, x.s)
 
 """
+    Interpolator(s)
+
+An iterator that yields _interpolation chunks (_i.e._ a string if the chunk was a string literal and
+an `Expr` if the chunk was an interpolation).
+
+# Examples
+"""
+struct Interpolator{S<:AbstractString}
+    s::S
+end
+
+Base.iterate(
+    interp::Interpolator,
+    i=firstindex(interp.s)
+) = iterate_interpolations(interp.s, i)
+Base.IteratorSize(::Interpolator) = Base.SizeUnknown()
+
+"""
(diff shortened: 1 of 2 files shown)


travigd:
Thanks @piever !

>There's a basic trick that tends to make these things more readable which is the following. Instead of a manual `iterate_interpolations(s, i=firstindex(s))`, you could do:
```julia
struct Interpolator{S<:AbstractString}
  s::S
end
function Base.iterate(interp::Interpolator, i = firstindex(interp.s))
  s = interp.s
  .... # same code body as you have
end
```
> and then basically you can do `[f(x) for x in Interpolator(str)]` and things like that
or say `Iterators.filter(filter_func, Interpolator(str))` and so on (basically use every function that works on iterators)

src/syntax.jl
```diff
@@ -55,27 +55,100 @@ end
...
+An iterator that yields _interpolation chunks (_i.e._ a string if the chunk was a string literal and
+an `Expr` if the chunk was an interpolation).
+
+# Examples
+"""
+struct Interpolator{S<:AbstractString}
+    s::S
+end
+
+Base.iterate(
```
piever: I think this could just be
```julia
Base.iterate(interp::Interpolator, args...) = iterate_interpolations(interp.s, args...)
```

travigd:
It turns out string parsing is hard and Julia does strange™ things when quotes are involved. But I **think** it's good to go now (and @piever your suggestions were incorporated, thank you!).

PR #257, opened 2019-03-29
Working and usable RPCs

### LINKED ISSUE #216: Remote procedure calls in WebIO

travigd:
Toy example. Closes #216.

```julia
using WebIO
x = Ref{Any}(nothing)

my_rpc = WebIO.RPC() do newval
    x[] = newval
end

w = Scope()
onclick = js"""
async function() {
    console.log(this);
    const value = this.parentElement.querySelector('input').value;
    await $my_rpc(value);
}
"""
w(dom"div"(
    dom"input"(),
    dom"button"(
        "update",
        events=Dict("click" => onclick),
    ),
))
```


travigd:
This PR also fixes the Lerna setup (for good this time!). It's unfortunate that it's bundled with this PR but it had to be done to fix CI.

src/rpc.jl
```diff
@@ -1,43 +1,62 @@
...
         rpc
     end
 end
 
 (rpc::RPC)(args...; kwargs...) = rpc.func(args...; kwargs...)
 
-rpcs = Dict{String, RPC}()
+registered_rpcs = Dict{String, RPC}()
+
+tojs(rpc::RPC) = js"WebIO.getRPC($(rpc.id))"
```
shashi: Why can't we just do this for any function that gets passed to `tojs`? Why do we need the RPC type?

travigd:
I think we could do that. Convert it to generic code that sends the arguments and have Julia try to look up a method by that name. I think I initially thought it was a bit "magic" and was planning to leave that kind of thing to JSExpr, but I can write a PR that updates it to that.

PR #259, opened 2019-03-29
reduce piracy to Widgets

piever:
The idea with the current Widgets design is that all "graphical things" just forward stuff to `Widgets.render(x)` (separate from `WebIO.render`), so that packages different than WebIO can use `Widgets` (the plan is to add a Makie backend: https://github.com/JuliaPlots/AbstractPlotting.jl/pull/88). So basically for a Makie `Widget`, `Widgets.render(wdg)` would return a Makie object that would then be displayed with the Makie pipeline. The Mux and Blink methods should live in Mux and Blink to avoid type piracy but I thought we can move them when we move the rest.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,4 +5,4 @@ Observables 0.2.2
 AssetRegistry
 Requires
 Compat 0.59
-Widgets 0.3.1
+Widgets 0.5.1
--- a/src/node.jl
+++ b/src/node.jl
@@ -142,7 +142,6 @@ end
 
 Base.show(io::IO, m::MIME"text/html", x::Observable) = show(io, m, WebIO.render(x))
 Base.show(io::IO, m::WEBIO_NODE_MIME, x::Union{Observable, AbstractWidget}) = show(io, m, WebIO.render(x))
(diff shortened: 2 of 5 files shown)


shashi:
That's pretty cool! So in theory after #5 WebIO will not need to depend on Widgets.jl?

piever:
That's an interesting point of view. In theory I'd say yes, even though in practice there is the issue that Widgets also defines the placeholder `node` so that various packages could do layout (using `node` and some css strings) just depending on Widgets: this is clear type piracy though. I imagine that the solution would be to apply the same trick and define layout primitives in Widgets as well (say `hbox` and `vbox`) which would call `hbox(get_backend(), args...)` which in turn would be implemented by various packages (so `CSSUtil` would implement it for Interact backend, AbstractPlotting for the Makie backend). @SimonDanisch was planning to do this in a LayoutBase package (rather than Widgets), I'm not sure whether it makes a big difference (maybe we could just rename Widgets to UIBase and have the two things together, I find it confusing when there are too many packages for related things).

shashi:
I'm not sure how useful this kind of abstraction will continue to be, I think we should not get caught up in it.

PR #262, opened 2019-03-30
Simplify RPC API.

travigd:
Simplifies RPC API. You can now interpolate any Julia function.

```julia
js"""
async function() {
    // Note: without `await`, this executes asynchronously.
    // The println **should** start execution before the abs, but the order of completion
    // is absolutely not guaranteed (except in IJulia, due to the fact that comms are
    // synchronous).
    $println("Print to Julia's stdout.");
    console.log(await $abs(-3));
}
"""
```

Equivalently, using `Promise.then` rathern than `async` functions, you can write this.
```julia
js"""
function() {
    $abs(-3).then((result) => console.log(result));
}
```

Requesting review from @shashi.

### DIFF
--- a/src/rpc.jl
+++ b/src/rpc.jl
@@ -1,44 +1,12 @@
-export RPC
+# A map from `hash(f)` to `f`.
+# This is used to lookup the function when we get a request from the frontend.
+registered_rpcs = Dict{UInt, Function}()
 
-"""
-    RPC(function; <keyword arguments>)
-
-A function wrapper that can be used to allow Javascript code to call
-    user-defined Julia functions.
-
-# Examples
-```julia
-my_add = RPC((x, y) -> x + y)
-my_mult = RPC() do x, y
-    x * y
+function tojs(f::Function)
+    h = hash(f)
+    registered_rpcs[h] = f
+    return js"WebIO.getRPC($(string(h)))"
 end
-my_abs = RPC(abs)
-
-# Create an onClick event handler.
-onclick = js\"""
-async function handleClick(event) {
-    console.log(await \$my_add(1, 2), await \$my_mult(3, 4), await \$my_abs(-5));
-}
-\"""
-dom"button"("Use RPCs!", events=Dict("click" => onclick))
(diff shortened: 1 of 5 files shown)


src/rpc.jl
```diff
@@ -1,45 +1,13 @@
...
-    id::String
-
-    function RPC(func::Function; id::String=newid("rpc"))
-        rpc = new(func, id)
-        registered_rpcs[id] = rpc
-        rpc
-    end
+# A map from `hash(f)` to `f`.
+# This is used to lookup the function when we get a request from the frontend.
+registered_rpcs = Dict{UInt, Function}()
```
shashi: Generally anything can be made callable in Julia... So just make this Any, the `hash` should still work.

shashi:
Excellent!

Did you try this out with JSExpr?

An idea for a `@jl` macro you might enjoy:

```
@js () -> @jl println("callback was called")
```
Translates to
```
code = () -> println("callback was called")
js"function () { $code() }"
```

to be clear, this should be in JSExpr.

src/rpc.jl
travigd: Sure, but people would have to write their own `tojs` functions for that to actually work, I think.

travigd:
I think we could even have it translate to
```julia
js"""function() { $(() -> println("callback was called")) }"""
```
since that's valid.

Question: Should we `await` by default? At the very least we need to make it very clear what the semantics are.

src/rpc.jl
shashi: Good point. Looks good then. :)

shashi:
Well is there a way to asynchronously call in that case?

in both the es6 and older cases?

If not, then the what we have seems more general.

Also it's very common for JS libraries to have Async APIs in cases like this right?

travigd:
I'm sorry, I don't think that was clear. I meant in JSExpr.jl for that last question. But we can have that discussion in that repo (opening an issue there now...).

At any rate, I think this can be merged before we have that in.

PR #263, opened 2019-03-30
Re-work IFrame handling.

### LINKED ISSUE #225: IFrame scope issues in 0.6.0

travigd:
Fixes #225.

This works by loading the generic HTTP provider JS code inside of the IFrames. It does have one **major** caveat though, which is that the iframes never actually go away. I think this was an issue before.

Basically, we setup the "child" WebIO instance to forward messages to and from the "parent" WebIO instance. Unfortunately, there's no real way to destroy that relationship (and thus allow the "child" WebIO to be garbage collected) when the IFrame is destroyed (or, at the very least, I couldn't make it work). The issue is that there's no "I was removed from the DOM" event.

This is mostly only an issue with the IJulia provider since that's where we're likely to create and destroy lots of IFrames.

I don't think this is a regression from WebIO 0.5.0, but it's definitely less than ideal in terms of memory utilization. Generally, I think we should just caution that IFrames are to be used VERY sparingly.

Ping @rdeits


rdeits:
This sounds great, thanks! I'm perfectly happy to avoid creating iframes whenever possible, but I think it will be be extremely useful to have them available, at least in limited numbers.

PR #271, opened 2019-04-15
fix webserver for HTTP 0.8 + WebSockets 1.5

SimonDanisch:


pfitzseb:
Re-add the [tests](https://github.com/JuliaGizmos/WebIO.jl/blob/master/test/http-tests.jl#L6)? :)

PR #280, opened 2019-05-03
improve http header

SimonDanisch:


src/providers/generic_http.jl
```diff
@@ -16,16 +16,28 @@ const bundle_key = AssetRegistry.register(normpath(joinpath(
...
 Serve an asset from the asset registry.
 """
 function serve_assets(req)
     if haskey(AssetRegistry.registry, req.target)
         filepath = AssetRegistry.registry[req.target]
         if isfile(filepath)
+            mime = get(
+                known_mimetypes,
+                extension(filepath),
+                "application/octet-stream"
```
pfitzseb: Is this really a good fallback? Or could we just not send a `"Content-Type"` if it can't be determined?

src/providers/generic_http.jl
```diff
@@ -16,16 +16,28 @@ const bundle_key = AssetRegistry.register(normpath(joinpath(
...
+            )
             return HTTP.Response(
                 200,
-                [],
+                [
+                    "Access-Control-Allow-Origin" => "*",
```
pfitzseb: AFAIU it's a "best practice" not to set this to `"*"`, but  ¯\\\_(ツ)\_/¯ I guess :P

SimonDanisch: Oh... well maybe it's not needed anyways :D

travigd: Actually, I'm not sure that this is necessary at all. `GET` requests that don't include [strange headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests) shouldn't trigger CORS at all.

src/providers/generic_http.jl
```diff
@@ -16,16 +16,27 @@ const bundle_key = AssetRegistry.register(normpath(joinpath(
...
+const requests = []
 """
 Serve an asset from the asset registry.
 """
 function serve_assets(req)
+    push!(requests, req)
     if haskey(AssetRegistry.registry, req.target)
         filepath = AssetRegistry.registry[req.target]
         if isfile(filepath)
+            headers = ["Access-Control-Allow-Origin" => "*"]
```
travigd: I think this section would read better if you defined a `filename2mime` style method (or `ext2mime` or whatever) and then simply defined headers as
```
headers = [
    "Access-Control-Allow-Origin" => "*",
    "Content-Type" => filepath2mime(filepath),
]
```

Not a deal breaker either way.

src/providers/generic_http.jl
travigd: If CORS is being triggered, then maybe SystemJS adds strange things to the headers that aren't allowed under the specification for _simple requests_.

src/providers/generic_http.jl
travigd: `application/octet-stream` is a good fallback IMO. It's the default (according to the HTTP spec) if the content type isn't specified and can't be guessed (by inspecting the file and using whatever heuristics the browser has implemented).

Source: http://mattryall.net/blog/2008/03/default-content-type

src/providers/generic_http.jl
travigd: Oh! Sorry for the spam, I was mistaken. _Simple requests_ don't trigger **preflight** requests (_i.e._ `OPTIONS` requests that precede the real request), but they still subscribe to CORS rules about whether or not the client (the originator of the XHR) can access the content.

tl;dr: If we're only using (_simple_) `GET`'s then we **don't** need to worry about `OPTIONS` requests but **do** need to set the `Access-Control-Allow-Origin` header (and I think in this case, `*` is acceptable).

So what @SimonDanisch has is correct.

src/providers/generic_http.jl
pfitzseb: Ah okay, didn't I kinda assumed HTTP would default to something else. Carry on :P

PR #283, opened 2019-05-09
WIP: Respect props on Scopes

shashi:
My goal is to set `display: inline` on the wrapper node so that I can have inline LaTeX which are Scope.

I don't at all think this is the right way to do it, but @travigd need your help.

I get this on build:

```
src/Scope.ts(127,16): error TS2339: Property 'props' does not exist on type 'ScopeSchema'.                                                                                         
src/Scope.ts(128,20): error TS2339: Property 'applyProps' does not exist on type 'typeof WebIODomNode'.                                                                            
src/Scope.ts(128,50): error TS2339: Property 'props' does not exist on type 'ScopeSchema'.
```

I am not sure if I should add the props field to ScopeSchema itself or do a ScopeDataSchema like you do for Node. Also not sure why applyProps is not defined on the WebIODomNode class.

### DIFF
--- a/packages/package-lock.json
+++ b/packages/package-lock.json
@@ -3834,7 +3834,7 @@
 		},
 		"duplexer": {
 			"version": "0.1.1",
-			"resolved": "https://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
+			"resolved": "http://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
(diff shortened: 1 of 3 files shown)


pfitzseb:
I really want this for TableView as well, so :+1:.

travigd:
Sorry for letting this slip past me for so long. I'd like to do merge #312 instead and keep the separation between node and scope. Let's continue the conversation there.

PR #284, opened 2019-05-15
switch to CBOR

SimonDanisch:
CBOR is around 10x faster than JSON (see https://github.com/MikeInnes/BSON.jl/pull/38) and can reduce size by a factor of ~2x.

This PR changes the websocket connection to use CBOR and should work fairly well already (WGLMakie runs without problems, which should use most of WebIOs features).

 TODOs:
- [ ] switch json lowering to CBOR
- [ ] make the encoding/decoding backend easier to switch out, so that we can more easily benchmark and switch to whatever the best format is more quickly 
- [ ] maybe improve WebSocket to directly encode/decode the stream of data without intermediates (might actually not be faster, depending on how the websocket stream works)
- [ ] alternatively, use a pre allocated buffer for decoding/encoding


rdeits:
This looks like it would be beneficial for MeshCat.jl too. We're currently doing some unnecessary base64 encoding in order to fit into json, but I'd love to just send byte arrays directly.

travigd:
We probably won't be able to get around base64 in general (it's required for Jupyter afaik - though there might be a way around that using the message buffers, though I'm unsure if that's available with IJulia or the Jupyter notebook client).

We could have the conversion from binary to base64 happen in WebIO, dependent upon the provider, but that's somewhat complicated.

SimonDanisch:
I was thinking maybe we should just drop the whole plain text lowering, and simply show an empty webio instance, and sent it a message with the data it should append to the DOM...
Or the other way around, have it get the data like it's done with RPCs.
I know there is also the use case to package something into one string and ship that, but I feel that we should have clearer APIs for those different use cases anyways.

shashi:
What's the strategy with Jupyter? It's unclear from the conversation so far...

SimonDanisch:
None :D So far I had 2 strategies in my mind: any provider can use their own serialization protocol, so Jupyter could just continue doing what it does... Or base64 encode this.
Although, I must say, CBOR won't be it...JS support is terrible, and it will make more sense to use MessagePack, or actually use JSON3 & compression, which could solve most of the performance problems as well - since JSON de/encoding in JS is relatively fast compared to other frameworks, mainly because it seems nobody has really figured out how to write fast serialization frameworks in JS :D 
Some benchmarks: https://nextjournal.com/a/LDdRcMcrQcPVe6Rj5SUWR?token=Nf4Tw3asM9tdauSGrPFXZJ

travigd:
Closing as stale. We talked about JSON3 at JuliaCon.

Issue #279, opened 2019-04-30
deps/build.jl creates %APPDATA% folder in my home directory on Windows


alkorang:
![image](https://user-images.githubusercontent.com/11420879/56961341-8278e580-6b8e-11e9-971d-7983fe20d98c.png)
540848e55c47eb13a189b085f2994ed6694b8727
Maybe use `joinpath(ENV["%APPDATA%"], "jupyter")` instead?

Edit: No, `EVN["APPDATA"]` without `%`

travigd:
We should use IJulia's method of discovering APPDATA since it _works_.
https://github.com/JuliaLang/IJulia.jl/blob/master/deps/kspec.jl

travigd:
We should probably just make IJulia a hard dependency (as opposed to the "soft" dependencies like Blink and Mux) and use their implementation of `jupyter_data_dir()`.

shashi:
That would be bad for apps that just want to use Blink (@ranjanan has one)

travigd:
How about just checking if IJulia is installed in the build process and only installing the extension then? The trade-off is that if IJulia was installed later, the extension wouldn't be installed, but we could move the installation process into a function so people could call `WebIO.install_nbextension()` (and include that bit of help in the `WebIO not detected` message, or link to maybe a section of the readme with troubleshooting information).

How does that sound?

pfitzseb:
We could also check via Requires if IJulia and WebIO are loaded in the same Julia process, no? And only install the nbextension if that's the case.

travigd:
Yes but I don't think trying to install the extension every time WebIO is loaded is the best thing to do. For example if someone (for some reason) wanted to remove the extension we shouldn't force them to install it every time. Also, in the case of the notebook, it won't work since the JavaScript wouldn't be loaded until refresh anyway.

travigd:
Closed by #286.

Issue #160, opened 2018-07-23
no output on MacOS


AshtonSBradley:
I am using Jupyter on MacOS (in Chrome) and getting no output. 

If I run

```julia
using WebIO
display(Node(:input, attributes = Dict("type" => "range")))
```
nothing appears.

```julia
Pkg.status("WebIO")
 - WebIO                         0.2.4
```

```julia
versioninfo()
Julia Version 0.6.4
Commit 9d11f62bcb (2018-07-09 19:09 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell MAX_THREADS=16)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)
```

SimonDanisch:
I think I can reproduce and see errors in the console:
![grafik](https://user-images.githubusercontent.com/1010467/43107463-6a75a88a-8edd-11e8-9031-70e61653b432.png)

SimonDanisch:
Hm, I tried 0.2.2 up to master and it all didn't work^^ Also 0.2.5 + 0.2.6 seem to be unknown to metadata?!
Anything I'm missing, or has the IJulia support been broken for a long time now? ;)

AshtonSBradley:
ah, yes, that would be consistent with an issue I opened on InteractNext.jl a while back ( I didn't realise that everything had moved over to InteractBase)

AshtonSBradley:
I am happy to try this. Just to check - should I be clearing .julia first, or can I initialise a package repository somewhere else?

rdeits:
No, you don't need to clear out `.julia`. To make a nice clean package directory, you can do the following: 

* (in your terminal, make a new temporary directory and `cd` into it
* Create a new package folder with `mkdir packages` 
* Tell Julia to use it by setting the environment variable with `export JULIA_PKGDIR=`pwd`/packages`
* Start Julia
* `julia> Pkg.init()`
* `julia> Pkg.add("IJulia")`
* etc.

Julia will use that package directory whenever it's launched from a terminal with the `export JULIA_PKGDIR=...` set.

AshtonSBradley:
[...]

PR #286, opened 2019-05-18
Fix %APPDATA% directory on Windows (for real this time).

travigd:
Tested on Windows.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,15 +1,14 @@
 using JSON
 
+include("./jupyterdirs.jl")
+
 const BEGIN_MARKER = "###JULIA-WEBIO-CONFIG-BEGIN"
 const END_MARKER = "###JULIA-WEBIO-CONFIG-END"
-function install_ijulia_config()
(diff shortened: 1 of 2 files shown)


travigd:
Fixes #279.

travigd:
Ping @shashi as well. This removes the configuration in the JSON file (`jupyter_notebook_config.json`) -- the JSON file shouldn't(?) work anyway because it won't have the location of the `jlstaticserve.py` in `sys.path`.

Was there a reason for adding that in the first place or am I missing something?

travigd:
Oh, I see: https://github.com/JuliaGizmos/WebIO.jl/issues/160 is relevant here.

I think the fix is to add to `sys.path` in the `.py` file and add the `nbserver_extensions` bit only in the JSON. I will make that change.

travigd:
@rdeits @SimonDanisch 
You're recommended by GitHub so you're being dragged into this. Can I get a pair of eyes on this before merging?

PR #291, opened 2019-05-26
Refactor file structure.

### LINKED ISSUE #289: Refactor file layout.
There are currently some weird things with circular relationships. We should try to build a more _layered_ approach. This will also make it more obvious where things should go, give us a chance to clean up some cruft, and just make WebIO a bit more pleasurable to hack on.

By _layered_ I just mean having more rigid layers of one-way abstraction. For example, `Scope` is probably the highest level of abstraction so other files shouldn't reference scopes. For example, the `import!` method currently exists in `imports.jl` (which definitely makes sense) but it acts on a `Scope` and so should belong in `scope.jl` (because the imports are a lower level of abstraction -- scopes should depends on imports and not the other way around).

travigd:
Closes #289.

There are some other things that we could tackle either in this PR or other ones. The initial commit only moves code around and doesn't add or remove anything.

## Other things to address
- [x] Why is `descendants_count` (in `Node`) a thing?? It doesn't seem to be used anywhere.
- [ ] ~What is the point of using FunctionalCollections.jl? (Not sure this is something to be *fixed* per se, just unclear as to what the purpose is.)~

Pinging basically everyone: @shashi @SimonDanisch @rdeits @piever @pfitzseb.

### DIFF
--- a/src/WebIO.jl
+++ b/src/WebIO.jl
@@ -14,6 +14,19 @@ The filesystem path where the WebIO frontend packages lives.
 """
 const packagepath = normpath(joinpath(@__DIR__, "..", "packages"))
 
+"""
+The MIME type for WebIO nodes.
+
+This is used when serializing a WebIO node tree to a frontend.
+For example, when using Jupyter, the frontend will render the serialized JSON
+via this MIME bundle rather than using the data sent as `text/html`.
+"""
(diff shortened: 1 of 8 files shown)


travigd:
This last commit (4150619d1b5be72316c6ff5fb8a74ff005b6c441) goes a bit beyond the scope of just moving files around so if we'd rather I can revert that and do that kind of thing in separate PR's.

PR #292, opened 2019-05-27
[WIP] Enhance documentation.

travigd:
The README is a bit aged at this point. It should also probably be split into separate documentation with only the bare minimum contained in the README (possibly using some other tool like Documenter.jl).

As of initial commit, I've updated about half of it.


travigd:
I added [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) documentation. It's a good start. I don't think we need to get everything perfectly documented before merging this, but would appreciate some feedback.

To build locally, just `cd ./docs` and `julia ./make.jl`. Works best to run a webserver to serve the documentation: `cd ./build && python3.6 -m http.server`.

PR #295, opened 2019-05-29
Deprecate custom scope ids

SimonDanisch:
The scope ID should 100% internal, and no one should be able to mutate it... Or use it to name the scope. 
This change should be making sure of this. 
If packages actually used the ID to name the scopes, we should give an alternate option to name scopes, and not mix it with the ID.

### DIFF
--- a/src/scope.jl
+++ b/src/scope.jl
@@ -17,27 +17,65 @@ export Scope,
 import Compat.Sockets: send
 import Observables: Observable, AbstractObservable, listeners
 
+# bool marks if it is synced
+const ObsDict = Dict{String, Tuple{AbstractObservable, Union{Nothing,Bool}}}
+
 mutable struct Scope
-    id::AbstractString
     dom::Any
-    observs::Dict{String, Tuple{AbstractObservable, Union{Nothing,Bool}}} # bool marks if it is synced
+    observs::ObsDict
     private_obs::Set{String}
-    systemjs_options
+    systemjs_options::Any
     # This should be an array of scopes, but there currently exist some circular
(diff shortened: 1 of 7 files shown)


src/scope.jl
```diff
@@ -17,27 +17,65 @@ export Scope,
...
+        return scope
+    end
+end
+
+# Need to use String for correct json javascript serialization -.-
+scopeid(x::Scope) = string(objectid(x))
+
+const global_scope_registry = Dict{String, Scope}()
+
+function register!(scope::Scope)
```
travigd: `register!` is a very generic name - do we want to use `register_scope!` instead? This would match `lookup_scope` as well.

I know that part of the generic-ness of the name `register!` is resolved by the fact that we dispatch on the `Scope` type, but it's not clear in the constructor what exactly `register!` is supposed to do.

src/scope.jl
```diff
@@ -17,27 +17,65 @@ export Scope,
...
+    global_scope_registry[scopeid(scope)] = scope
+end
+
+function deregister!(scope::Scope)
```
travigd: Same comment as above - `deregister_scope!` might be better.

travigd:
This is also technically a breaking change so we need to deprecate it properly.

Something like this.

```julia
function Base.getproperty(scope::Scope, property::Symbol)
    if s == :id
        Base.depwarn("Accessing scope.id is deprecated; use scopeid(scope) instead.", :webio_scope_id)
        return scopeid(scope)
    end
    return getfield(scope, property)
end
```

PR #298, opened 2019-05-30
README: Just link directly to the Interact.jl repo…

mbauman:
and not its issues page


travigd:
CI fail is unrelated (typescript update just released is what I think is causing it).

Thank you! I noticed this the other day - think it was because I just copied and pasted from my open tab.

PR #306, opened 2019-06-07
Add @register_renderable macro.

travigd:
This addresses some of the issues that currently exist around extending WebIO.

It allows for a library author to write
```julia
struct ScatterPlot
    x::Vector{Float64}
    y::Vector{Float64}
end

# All of the following are equivalent
# Option A
WebIO.@register_renderable function WebIO.render(x::ScatterPlot)
    # Construct the scatter plot using DOM primitives...
    return node(...)
end

# Option B
WebIO.@register_renderable WebIO.render(x::ScatterPlot) = node(...)

# Option C
function WebIO.render(x::ScatterPlot)
    ...
end
WebIO.@register_renderable ScatterPlot
```

This will automagically create the requisite `Base.show` methods for `WEBIO_NODE_MIME` and `text/html` which is required to work correctly in Jupyter.

Would help address https://github.com/sglyon/PlotlyJS.jl/issues/278.


piever:
I confess that I think there is some redundancy here... Once we have `register_renderable`, I imagine we don't really need the macro and `register_renderable` could also add a show method for `WEBIO_NODE_MIME`? Why defining one show method is a function and defining two is a macro?

As a counter-proposal, regardless of whether we choose function or macro (I personally prefer a macro not to have a `@eval` inside a function body), I would propose two methods:

```julia
macro register_renderable(T)
    # define show methods for T
end
```
and
```julia
macro register_renderable(f, T)
    # define show methods for T
    # define WebIO.render(x::T) = f(x)
end
```
To be used with the syntax:
```julia
@register_renderable(MyType) do val
    # define corresponding node
end
```

And we could probably deprecate `register_renderable` as a function.

travigd:
Yes, the plan (after I realized that I duplicated some register renderable stuff) is definitely to deprecate the function version.

I also very much like that syntax (in addition to the bare `@register_renderable MyType`).

travigd:
@shashi @piever Are y'all happy with this?

There's a bug(?) in Julia that makes the my preferred syntax slightly impossible (see https://github.com/JuliaLang/julia/issues/32301).

```julia
# This doesn't work.
WebIO.@register_renderable(MyType) do x
    return node(:p, ...)
end
```

It does work with the uglier syntax where we put the `@` before the `WebIO.` qualification:
```julia
# This works
@WebIO.register_renderable(MyType) do x ... end
```

It also works if you directly import the name.
```julia
# This works
using WebIO: @register_renderable
@register_renderable(MyType) do x ... end)
```

src/render.jl
```diff
@@ -174,4 +157,196 @@ end
...
+        end
+        f_args = f_args.args[1]
+    end
+    if !isa(f_args, Expr) || f_args.head != :(::)
+        # Add type annotation so that we define WebIO.render(x::MyType)
+        f_args = Expr(:(::), f_args, typename)
+    end
+    f_body = f.args[2]
+
+    render_method_expr = Expr(
```
piever: All this code seems a bit to complicated, wouldn't it be easier to just do:

```julia
quote
    local func = $funcname
    WebIO.render(arg::$typename) = func(arg)
end
```

PR #307, opened 2019-06-08
Deprecate blink provider.

travigd:
Counterpart to https://github.com/JunoLab/Blink.jl/pull/201.

If/when we merge this, we should have a deprecation period and then commit to a certain time to pull the plug on the Blink provider (possibly with other breaking changes). Possibly mark that release as 1.0.


src/WebIO.jl
```diff
@@ -14,6 +14,11 @@ The filesystem path where the WebIO frontend packages lives.
 """
 const packagepath = normpath(joinpath(@__DIR__, "..", "packages"))
 
+"""
+The path of the main WebIO JavaScript bundle file.
+"""
+const bundlepath = normpath(joinpath(packagepath, "webio", "dist", "index.js"))
```
travigd: Technically speaking, this is not a bundle. We could set up webpack and ship it as a bundle, though. It would make more sense for something like Mux than it does here, though (it might actually be required for Mux, come to think of it).

PR #312, opened 2019-06-15
Make Scope wrapper element a span, not div.

travigd:
Unrelated, it'd be nice to fix how we deal with JS packages (I'll make an issue).


travigd:
Travis (the CI platform - not the very handsome author of this PR) seems to have failed due to a network issue. Retrying...

shashi:
Interesting.

On #283 you said

> keep the separation between node and scope

I think it's crucial that Node and Scope have the same API because:

```julia
something_rendered = NotMyPackage.showsomething(something)

node(:div,
   style(something_rendered, :display=>:inline),
   mything())
```

Should work without knowing which type `NotMyPackage.showsomething` returns. It's just the correct modularity. So I prefer the #283 style fix!

shashi:
Also this change here is breaking. It breaks all code that assumed scopes are divs so far.

travigd:
> Also this change here is breaking. It breaks all code that assumed scopes are divs so far.

I'll take things that aren't public API but people have definitely relied upon for 500, Alex. (Ugh.)

What does `style(::Scope, ...)` even do right now? I'd be in favor of keeping this change (we can change it to a div with `display: inline`) and define `style(s::Scope, args...) = style(s.dom, args...)`. Adding all of the normal DOM stuff to Scope is gonna be a nightmare imo.

shashi:
> define `style(s::Scope, args...) = style(s.dom, args...)`. 

Ya.

Ideally though we should get rid of an extra wrapper element altogether, I can't remember what exactly was the problem with that though... https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment will allow us to use a fragment instead of a div or span...

travigd:
Closing because don't want to deal with rebasing.

Issue #317, opened 2019-07-02
latest release (v0.8.6) broke Blink master


NHDaly:
## The bug

It looks like the latest release of WebIO to `v0.8.6` (https://github.com/JuliaRegistries/General/pull/1473) broke Blink on master:
https://travis-ci.org/JunoLab/Blink.jl/builds/553312986
```julia
ERROR: LoadError: Javascript error	Error: Cannot find module 'debug'
Stacktrace:
 [1] #js#3(::Bool, ::Function, ::Page, ::WebIO.JSString) at /Users/travis/build/JunoLab/Blink.jl/src/rpc/rpc.jl:53
 [2] (::getfield(Blink, Symbol("#kw##js")))(::NamedTuple{(:callback,),Tuple{Bool}}, ::typeof(js), ::Page, ::WebIO.JSString) at ./none:0
 [3] #js#23(::Bool, ::Function, ::Window, ::WebIO.JSString) at /Users/travis/build/JunoLab/Blink.jl/src/AtomShell/window.jl:213
 [4] #js at ./none:0 [inlined]
 [5] #js#4(::Bool, ::Function, ::Window, ::Expr) at /Users/travis/build/JunoLab/Blink.jl/src/rpc/rpc.jl:61
 [6] js at /Users/travis/build/JunoLab/Blink.jl/src/rpc/rpc.jl:61 [inlined]
 [7] initwebio!(::Window) at /Users/travis/build/JunoLab/Blink.jl/src/AtomShell/webio.jl:16
 [8] #Window#17(::Bool, ::Type, ::Electron, ::Page, ::Dict{Any,Any}) at /Users/travis/build/JunoLab/Blink.jl/src/AtomShell/window.jl:70
 [9] Type at ./none:0 [inlined]
 [10] #Window#16(::Bool, ::Type, ::Electron, ::Dict{Any,Any}) at /Users/travis/build/JunoLab/Blink.jl/src/AtomShell/window.jl:44
 [11] (::getfield(Core, Symbol("#kw#Type")))(::NamedTuple{(:async,),Tuple{Bool}}, ::Type{Window}, ::Electron, ::Dict{Any,Any}) at ./none:0
[...]

Issue #315, opened 2019-07-01
WebIO not detected in jupyterlab


AshtonSBradley:
## The bug

When using `Interact.jl` in jupyterlab, specifically `@manipulate`, one gets the message
"WebIO not detected."
and no output is produced. 
<img width="737" alt="Screen Shot 2019-07-01 at 12 19 44 PM" src="https://user-images.githubusercontent.com/24194258/60404195-9e365180-9bfa-11e9-9242-dca0c6e06c4e.png">
## Context

Your context here.
```julia
julia> Pkg.status("WebIO")
    Status `~/.julia/environments/v1.1/Project.toml`
  [0f1e0344] WebIO v0.8.5

julia> versioninfo()
Julia Version 1.1.0
Commit 80516ca202 (2019-01-21 21:24 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 8
```

Using `jupyterlab`. I don't know how to run `jupyter labextension list` for the containerized `jupyterlab` installed by `IJulia`. Possibly I am just missing an extension? 

It works fine in the `jupyter` notebook from `IJulia`.

travigd:
Do you have more than one version of Jupyter Lab (e.g. one on your system and the one installed by IJulia)?

AshtonSBradley:
yes and no: I get the same error on my laptop where I only have `IJula`'s version, and on my desktop where I also installed anaconda to see if that would fix it. If I open Jupiterlab from the terminal using anaconda, same result

travigd:
Can you `Pkg.build("WebIO")` and see if that fixes things? If not, copy the output here?

We try to find the best JupyterLab to use; it defaults to the one in your path and if there is none it tries to use the one IJulia installs (if any).

travigd:
Can you `Pkg.build("WebIO")` and see if that fixes things? If not, copy the output here?

We try to find the best JupyterLab to use; it defaults to the one in your path and if there is none it tries to use the one IJulia installs (if any).

AshtonSBradley:
I get identical result to screenshot above, after (including the jupyterlab() Info so you see which one is run)
```julia
(v1.1) pkg> build WebIO
[...]

Issue #314, opened 2019-06-20
Don't silently install things into other software systems


davidanthoff:
I installed the Juno integration for Atom the other day. And now, all of a sudden, I'm getting messages inside jupyterlab that I need to rebuild webio.

I generally think this is really not ideal: if I install something like Juno, the last thing I want to happen is that some package down in the stack silently without me opting in starts to install extensions into other things like jupyterlab.

travigd:
This is fair but also has some cons.

It requires extra steps for people who want to use Interact (for example) in Jupyter or JupyterLab to go through an extra step. Things like IJulia also do this (writing to your local jupyter config directory) so it's not unheard of (though IJulia is also much more closely integrated with Jupyter than we are).

It might be good to surface a message when you run `using WebIO` when IJulia is loaded and it detects that it isn't installed (or just surfacing a message via HTML in the notebook) with instructions.

travigd:
I'm addressing this in #319.

I think the best course of action is going to be to not install by default (since we're a common dependency) and surface information about how to easily install the extension when we detect that we're running in Jupyter.

travigd:
Closed by #319.

Issue #313, opened 2019-06-15
JavaScript Package Management


travigd:
Currently, we include "compiled" (_i.e._ transpiled and/or bundled) JavaScript in this repository. It's not ideal.

## What do we do now?
Currently (this hasn't been super consistent), we've discussed `.gitignore`ing the generated files and force-adding them whenever we do a release. This is to avoid having lots of PRs that change these files (it also avoids the possibility of merge conflicts).

#### Why do we do this?
Because we don't want to make people install Node and NPM on their own machines in order to build the JavaScript themselves.

#### What are the issues?
We're still including many relatively large files in our git history which makes the repository larger (and slower to download). Also, it's not very _ergonomic).

## What should we do instead?
I'd like to set up a CI script that will run whenever a new version of WebIO is tagged (this happens automagically with TagBot when we update the version and register with the General metadata repo). The CI pipeline would generate the transpiled/bundled stuff and upload to NPM (and/or GitHub release assets) and the build step of installing WebIO would download these files (by reading the version of WebIO in Project.toml or something similar).

#### Pros
* No more dealing with generated JS in the codebase.
* Simplifies JupyterLab extension (simply run `jupyter labextension install @webio/jupyter-lab-provider` to get it for the corresponding version of WebIO).

#### Cons
* We depend on a third-party solution (NPM). I don't think this is a big deal since NPM is relatively reliable and we basically already do.

SimonDanisch:
Could we upload it to npm AND as release artifacts, so that a normal build can just download them from github?

PR #319, opened 2019-07-07
[WIP] Unify Julia + NPM publishing.

travigd:
We shouldn't have to check built-in files to Git.

The easiest thing to do is to publish things to NPM and then download the current version when being installed.


travigd:
I think publishing from CI is not gonna be the best route, here.

Instead I'm working on a `publish.jl` script which takes care of building the JS, testing against it, setting the NPM package versions, pushing to NPM, and creating the tag.

The reason why doing it in CI is problematic is because of the assumptions that Lerna make (which make sense when Lerna is the only tool being used to version... but... alas...).

travigd:
Closes #320.
Closes #317.
Closes #315.
Closes #314.
Closes #313.

travigd:
Closes #264 (we always build before testing if checked out for development).

travigd:
The VM which runs Julia 0.7 on Mac is apparently very flakey (it's happened twice where we get errors when it tries to download stuff from Conda). Going to merge this as is.

PR #338, opened 2019-08-05
Fix sending commands to multiple frontends.

### LINKED ISSUE #337: Pushing an observable to multiple clients no longer works in WebIO 0.8.9
With WebIO 0.8.9, opening the same scope through multiple frontends results in only one frontend getting the observable updates. This is breaking the MeshCat.jl tests. 

Here is a MWE:

```julia
using WebIO, Mux, Blink, JSExpr

scope = Scope()

# updates to this update the UI
count = Observable(scope, "count", 1)

onjs(count, # listen on JavaScript
     JSExpr.@js x->this.dom.querySelector("#count").textContent = x)

scope.dom = dom"div"(
    dom"div#count"(string(count[])),
)

@async while true
    count[] += 1
    sleep(1)
end

WebIO.webio_serve(Mux.page("/", req -> scope), 8000)

w = Window()
body!(w, scope)
```

If you comment out the `body!(w, scope)` line and just open `localhost:8000` in your browser, then everything is fine (the counter updates once per second). Likewise, if you comment out the `webio_serve` line, then a `Blink` window opens up and shows the counter updating. But if you keep both lines in, then the Blink window updates but the `localhost:8000` page does not. 

This works correctly with WebIO 0.8.8 but fails with 0.8.9. 

Tested with Blink 0.12.0, Mux 0.7.0, JSExpr 0.5.0.

travigd:
This also reworks the way that new channels are added to a scope to work more
synchronously and avoid some race conditions.

Fixes #337 .

### DIFF
--- a/src/connection.jl
+++ b/src/connection.jl
@@ -5,13 +5,6 @@ export AbstractConnection
 abstract type AbstractConnection end
 
 """
-The number of connections that we allow to be in a "pending" state (_i.e._ that
-haven't been moved from a `ConnectionPool`'s `new_connections` set to its
-`connections` set).
-"""
-const PENDING_CONNECTION_LIMIT = 32
-
-"""
 The maximum number of messages to allow into the outbox.
 """
 const DEFAULT_OUTBOX_LIMIT = 32
@@ -27,7 +20,7 @@ are automatically removed from the pool.
 struct ConnectionPool
     outbox::Channel
     connections::Set{AbstractConnection}
-    new_connections::Channel{AbstractConnection}
+    condition::Condition
 end
 
 function ConnectionPool(
@@ -37,7 +30,7 @@ function ConnectionPool(
     pool = ConnectionPool(
         outbox,
         connections,
-        Channel{AbstractConnection}(PENDING_CONNECTION_LIMIT),
+        Condition(),
     )
 
     # Catch errors here, otherwise they are lost to the void.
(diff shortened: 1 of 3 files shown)


travigd:
Fixes #337.

rdeits:
Thanks! I can confirm that this fixes the original issue. I'll just need to update the `wait()` implementation in MeshCat to use the new connection pool.

src/connection.jl
```diff
@@ -70,14 +64,12 @@ current task until that is the case. Also processes incoming connections.
 """
 function ensure_connection(pool::ConnectionPool)
     if isempty(pool.connections)
-        wait(pool.new_connections)
-    end
-    while isready(pool.new_connections)
-        push!(pool.connections, take!(pool.new_connections))
+        wait(pool.condition)
```
shashi: Little tiny race condition possible here if a new condition arrives in between `isempty` and `wait`. This might need a lock, eventually.

test/multiple-connections.jl
```diff
@@ -0,0 +1,39 @@
...
+    value = (w) -> @js w document.querySelector("[data-testid=foobar]").innerText
+    @test value(w1) == "initial"
+    @test value(w2) == "initial"
+
+    obs[] = "updated"
+    # Sleep to allow the JS to execute.
+    sleep(0.25)
+    @test value(w1) == "updated"
+    @test value(w2) == "updated"
+end
```
shashi: sweet tests!

tkoolen:
Could we ask you guys for a new tag with this fix? CC:@rdeits.

Issue #343, opened 2019-08-28
If a scope reaches its outbox limit, everything fails™.


twavv:
Two steps forward, one step back...

## Background
* Every scope has a `ConnectionPool` (aside - this is a bad name for what this is) which encapsulates all of the frontends that that scope is connected to (usually just one browser window, but sometimes more).
* When a message is sent to a scope (e.g. an update to an observable), the message is added to the `ConnectionPool`'s outbox where it will be processed asynchronously (every `ConnectionPool` has a `@async while True ... end` loop that waits for the outbox to have a message ready).

#330 introduces behavior to limit the size of the outbox because it was causing a memory leak when the frontend was closed (e.g. if you open a Blink window to display a plot, then close it, any updates to the plot will be queued in the scope's outbox, which creates a memory leak) **and** it says that we only queue messages until the first connection exists (this is to allow initial messages to be processed correctly - but subsequent messages can usually be safely ignored because they're just observable updates most of the time).

## Issue
But now there's an issue in Interact where sliders will stop updating. The reason for this is because every(?) widget creates multiple scopes, only some of which are rendered to the frontend. This means that after a certain number of updates (the default is `DEFAULT_OUTBOX_LIMIT = 32`), everything breaks because the scope which was never rendered starts throwing when we try to `put!` to the outbox channel.

There are actually multiple issues.
* WebIO breaks if the outbox limit is reached (this is, in a vacuum, probably a **good** behavior - things failing silently is a recipe for disaster).
* Interact creates throwaway scopes.
* IJulia seems to brick itself if a comm handler errors.

## Solution
I don't think we should buffer any messages and get rid of the outbox altogether. There are two options with this approach as far as I can see.
* Don't buffer anything and make sending synchronous. This means `myobs[] = ...` won't return until the update has been sent to all the frontends. This might result in issues if frontends hang, but that's not **usually** an issue.
* Have a queue for each connection (not each scope) that is not bounded. This doesn't have the same issue as #330 because the queue only exists while the connection exists (dead connections in a `ConnectionPool` are dropped every time a new message is sent) and solves the issue above because the channel isn't bounded.

rdeits:
For what it's worth, overwriting observable outputs with new updates would break MeshCat.jl. The messages sent through the observables in MeshCat.jl are incremental updates to the visualizer state, since re-transmitting the entire state would be too expensive. 

[...]

PR #346, opened 2019-09-08
Use an unbounded channel for each connection instead of an outbox

rdeits:
Here's a potential fix for #343. It's my best guess at how to implement the second option listed under "Solution" in that issue. 

In this implementation, each connection has its own buffer of messages and its own long-running task which exists only to take messages from the buffer and send them to the connection, potentially deleting the buffer if the connection is closed. If a scope has no connections attached, then trying to send a message to it will block until a connection is made. 

I've only done basic testing on this, but I can give it some more attention if the general approach seems reasonable.

### DIFF
--- a/src/connection.jl
+++ b/src/connection.jl
@@ -5,11 +5,6 @@ export AbstractConnection
 abstract type AbstractConnection end
 
 """
-The maximum number of messages to allow into the outbox.
-"""
-const DEFAULT_OUTBOX_LIMIT = 32
-
-"""
     ConnectionPool([outbox[, connections]])
 
 Manages the distribution of messages from the `outbox` channel to a set of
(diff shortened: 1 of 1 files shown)


src/connection.jl
```diff
@@ -18,46 +13,56 @@ and sends each message to every connection in the pool. Any closed connections
...
-        )
-    end
-
     return pool
 end
 
 function addconnection!(pool::ConnectionPool, conn::AbstractConnection)
-    push!(pool.connections, conn)
-    notify(pool)
+    if !(conn in keys(pool.connections))
```
travigd: Two points.
* Do you need the `keys(...)`? Doesn't `key in dict` work?
* This should be
```julia
if conn in pool.connections
    return
end
# Rest of function
```
so that the whole function body isn't in an if statement.

src/connection.jl
```diff
@@ -18,46 +13,56 @@ and sends each message to every connection in the pool. Any closed connections
...
+        let outbox = Channel{Any}(Inf)
```
travigd: Is this let necessary?

src/connection.jl
```diff
@@ -18,46 +13,56 @@ and sends each message to every connection in the pool. Any closed connections
...
+                        end
+                    catch exc
+                        @error(
+                            "An error ocurred in the while processing messages from a ConnectionPool.",
+                            exception=exc,
+                        )
+                        break
+                    end
+                end
+                println("Deleting: $conn")
```
travigd: Delete

src/connection.jl
rdeits: No--that works in Python, but `in` in Julia checks if the (key, value) pair is in the dict. `x in keys(dict)` is the canonical way to do it (and does not do anything silly like create a copy of the keys). But yes, your second point makes sense.

PR #353, opened 2019-10-10
Introduce hooks for new and lost connections for WebIO.Scope

rdeits:
This is an alternative to #346 which is much less invasive to the structure of WebIO. With this change I can essentially implement a design like #346 entirely within an external package (MeshCat in this case), using my own set of channels to mediate communication with the clients or just synchronously sending to each connection. 

For example, with this PR I can define my own `Router` that implements #346 like so:

```julia
struct Router
    scope::Scope
    channels::Dict{AbstractConnection, Channel{Any}}

    function Router(scope::Scope)
        router = new(scope, Dict())
        for conn in scope.pool.connections
            _add_outbox!(router, conn)
        end
        WebIO.on_open_connection(scope) do conn
            _add_outbox!(router, conn)
        end
        WebIO.on_close_connection(scope) do conn
            delete!(router.channels, conn)
        end
        router
    end
end

function _add_outbox!(router::Router, conn::AbstractConnection)
    if !(conn in keys(router.channels))
        let outbox = Channel{Any}(Inf)
            router.channels[conn] = outbox
            @async begin
                while true
                    # Catch errors here, otherwise they are lost to the void.
                    try
                        msg = take!(outbox)
                        WebIO.send(conn, msg)
                    catch exc
                        println("Error:", exc)
                        break
                    end
                end
            end
        end
    end
end

function send(router::Router, msg)
    for channel in values(router.channels)
        put!(channel, msg)
    end
end
```

### DIFF
--- a/src/connection.jl
+++ b/src/connection.jl
@@ -21,6 +21,8 @@ struct ConnectionPool
     outbox::Channel
     connections::Set{AbstractConnection}
     condition::Condition
+    open_connection_callbacks::Vector{Any}
+    close_connection_callbacks::Vector{Any}
 end
 
 function ConnectionPool(
@@ -30,7 +32,7 @@ function ConnectionPool(
     pool = ConnectionPool(
         outbox,
         connections,
-        Condition(),
+        Condition(), [], []
     )
 
     # Catch errors here, otherwise they are lost to the void.
@@ -50,6 +52,16 @@ end
 function addconnection!(pool::ConnectionPool, conn::AbstractConnection)
     push!(pool.connections, conn)
     notify(pool)
+    for callback in pool.open_connection_callbacks
+        callback(conn)
+    end
+end
+
+function removeconnection!(pool::ConnectionPool, conn::AbstractConnection)
+    delete!(pool.connections, conn)
+    for callback in pool.close_connection_callbacks
+        callback(conn)
+    end
 end
 
 function Sockets.send(pool::ConnectionPool, msg)
@@ -115,7 +127,7 @@ function send_message(
             send(connection, msg)
         else
             @info "Connection is not open." connection
-            delete!(pool.connections, connection)
+            removeconnection!(pool, connection)
         end
     catch ex
         @error(
@@ -123,7 +135,7 @@ function send_message(
                 * "frontend:",
             exception=ex,
         )
-        delete!(pool.connections, connection)
+        removeconnection!(pool, connection)
     finally
         return nothing
     end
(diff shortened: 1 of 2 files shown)


travigd:
I'm okay with adding some callbacks like this, but, I'm not sure that this is the right way to solve this specific issue. In particular, I don't think users should be directly invoking `send` on a connection instance (also, there's no way for the frontend to intercept that even if Julia did that).

I think it would be good to make the existing request/response/command stuff somewhat more extensible so that you could do something like

```julia
# In Julia
WebIO.command(conn, "my-command", Dict("data" => "data"))
```

```js
# In the frontend
WebIO.registerCommandListener("my-command", async () => {
  // Do stuff
});
```

The questions around that are how does that interact with scopes and connections? Commands are an easy case because you don't have to worry about responses from multiple frontends (so a method could dispatch work given either a scope or a connection).

The other issue is just making sure that the command listener is registered on the frontend side before it's sent on the WebIO side. This is possible with some manual RPC synchronization (i.e., when the frontend code is loaded, it invokes an RPC that lets Julia know that things are set up).

rdeits:
That all makes sense, but I'm also trying to deal with the fact that MeshCat is now broken with WebIO master due to the various outbox limit changes, and there's no easy way for the package resolver to work out a working set of dependencies: https://github.com/rdeits/MeshCat.jl/issues/127 . So while I'm happy to help with a more extensible command framework, I really need to figure out a concrete solution to the problem at hand. 

Or, to put it another way, if we're going for an extensible command interface, what is the first step? I just don't feel like I have my head sufficiently wrapped around the structure of this package anymore in order to make progress.

travigd:
So, for the specific issues relating to the outbox limit and MeshCat, I think the issue actually is solvable without any PR because you can pass custom outboxes to scopes. However, this isn't ergonomic and isn't possible if you start using things like Interact which don't expose that level of granularity with respect to WebIO.

To fix the aforementioned ergonomic-ity problem, we could change `const DEFAULT_OUTBOX_LIMIT = 32` to `default_outbox_limit = Ref{Int}(32)` and MeshCat could set it in its module `__init__()`.

Does that sound like it would work?

I still think extending the command capability is going to be a good idea but not a must-do-this-right-now kind of thing.

rdeits:
Ah, that's a good point about passing a custom outbox--I think that might solve the immediate problem. I'm not fond of making the default outbox limit a mutable global: it's bound to cause issues as soon as more than one package thinks it can set that value. 

I'll try the custom outbox and see if that gets things working again.

travigd:
I'm not sure that it's *thaaaat* bad since the only real use-case is to increase it to infinity. And when meshcat is being used it's probably the only one using WebIO. But if you wanna go another route I'm happy with not adding it. 😁

PR #356, opened 2019-11-01
Update for JSExpr v1.0.

travigd:
This reverses the direction of the dependency between WebIO and JSExpr (WebIO now depends on JSExpr).


shashi:
I don't like the direction reversal, ideally they should be separate packages....!!

shashi:
I know sorry I should have commented earlier, missed this. We can try again later.

Also everything is broken if I move to JSExpr v1.0.1 (Interact + all its dependencies that use `@js`), I had to pin to v0.5.1 to get Interact to work....

travigd:
Hmmm I remember Interact working with the latest JSExpr stuff.

Also, they are separate packages? JSExpr is now completely independent of WebIO which is what I thought we wanted.

Issue #368, opened 2019-12-21
Roadmap to v1.0


travigd:
This issue is meant to keep track of what needs to do/has been done to release WebIO v1.0.

- [ ] Update to JSExpr v1 (#356)
- [ ] Synchronous observable updates (#367)
- [ ] Add renderable trait (#361, this might deserve more discussion)
- [ ] Improve serialization for IJulia (#351)
- [ ] Finalize extensible messaging protocol (#323)
- [ ] Implement scope reaping (#321)
- [ ] Implement Observable pull-on-mount behavior (#258)
- [ ] Remove deprecations

I think there will be some fairly large API changes. I'm not too worried about it because most users of WebIO are really just using Interact/PlotlyJS/MeshCat and we can make sure that WebIO works for those uses cases.

PR #371, opened 2019-12-22
[Next] WebIO v1.0

travigd:
See #368


shashi:
Separate PR for each feature?

travigd:
> Separate PR for each feature?

Yeah I was planning on that to make things manageable while still keeping master usable if we need to make other changes.

aminya:
This seems like a huge PR. 😄. 

I want to add some features from WebIO into my package [AcuteML](https://github.com/aminya/AcuteML.jl/issues/33). Things like "External frontend interfaces" and "JavaScript interface" can be used for AcuteML too.

I should observe this PR to see what changes.

shashi:
> Yeah I was planning on that to make things manageable while still keeping master usable if we need to make other changes.

Yes please!

PR #378, opened 2019-12-31
Make WebIO statically compilable (relocatable): don't run `include()`s in __init__

NHDaly:
If you statically compile an application that uses WebIO, it will attempt to run this Module __init__ at runtime, but of course those source files will no longer be present once the application has been downloaded to a user's computer.

Please see these sections from the PackageCompilerX for an overview of relocatability:
- https://kristofferc.github.io/PackageCompilerX.jl/dev/apps/#Relocatability-1
- https://kristofferc.github.io/PackageCompilerX.jl/dev/devdocs/relocatable_part_3/

(I made this code change months ago but never sent it for review, so i don't remember what the errors were. But rereading the diff, it seems reasonable. I'll rebase and update it now..)

### DIFF
--- a/src/WebIO.jl
+++ b/src/WebIO.jl
@@ -87,15 +87,21 @@ function setup(provider::Symbol)
 end
 setup(provider::AbstractString) = setup(Symbol(provider))
 
-code_mux          = read(joinpath(@__DIR__, "providers", "mux.jl"), String)
-code_blink        = read(joinpath(@__DIR__, "providers", "blink.jl"), String)
-code_ijulia       = read(joinpath(@__DIR__, "providers", "ijulia.jl"), String)
-code_generic_http = read(joinpath(@__DIR__, "providers", "generic_http.jl"), String)
+function prefetch_provider_file(basename)
+  filepath = joinpath(@__DIR__, "providers", basename)
+  code = read(filepath, String)
+  (file = filepath, code = code)
+end
+
+provider_mux          = prefetch_provider_file("mux.jl")
+provider_blink        = prefetch_provider_file("blink.jl")
+provider_ijulia       = prefetch_provider_file("ijulia.jl")
+provider_generic_http = prefetch_provider_file("generic_http.jl")
 
 function __init__()
     push!(Observables.addhandler_callbacks, WebIO.setup_comm)
     @require Mux="a975b10e-0019-58db-a62f-e48ff68538c9" begin
-        eval(code_mux)
+        include_string(@__MODULE__, provider_mux.code, provider_mux.file)
     end
     @require Blink="ad839575-38b3-5650-b840-f874b8c74a25" begin
         # The latest version of Blink defines their own WebIO integration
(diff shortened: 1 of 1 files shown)


NHDaly:
@travigd Can i request a review from you? 😁 

---

Updated to use `include_string` instead of `eval` so that the filenames are still preserved, exactly as if it was a normal include:
```julia
julia> which(Base.isopen, (WebIO.WSConnection,))
isopen(p::WebIO.WSConnection) in WebIO at /Users/nathan.daly/.julia/dev/WebIO/src/providers/generic_http.jl:13
```

travigd:
Does this work with frontends? The JavaScript is still just static files that we include on the JS side for the most part.

I'm not familiar enough with the package compilation to even know where to begin with answering that. 🤓

NHDaly:
> Does this work with frontends? The JavaScript is still just static files that we include on the JS side for the most part.

I don't fully understand the question, unfortunately. What do you mean "the frontends"?

Anyway, I think this shouldn't change anything for normal julia execution.. we're just preloading the strings from the files during precompilation, and evaling them at runtime, instead of reading the files at runtime. This is needed because for a "shipped" executable, those files are no longer present at runtime.

But other than that, there should be no change! :)

travigd:
(preamble: I know nothing about the package compilation stuff so these might be rather basic questions and/or represent a faulty understanding of how things work)

My concern (or uncertainty, I guess) is around how the package compilation process will handle asset files.

For example, in Blink, we have some static JS and HTML and CSS files and they're references at runtime. For example, whenever a window is loaded, we `require` the WebIO JS bundle, which lives at `~/.julia/packages/WebIO/packages/webio/dist/webio.js` (or something like that). We use the pretty standard (afaict) way of referencing them:

https://github.com/JuliaGizmos/WebIO.jl/blob/e7585c59e90dcc078a8d11c01014f85a55f1e9eb/src/WebIO.jl#L17

My question is whether or not **that** is going to cause issues when we try to load it.

src/WebIO.jl
```diff
@@ -87,10 +87,21 @@ function setup(provider::Symbol)
 end
 setup(provider::AbstractString) = setup(Symbol(provider))
 
+function prefetch_provider_file(basename)
+  filepath = joinpath(@__DIR__, "providers", basename)
+  code = read(filepath, String)
+  (file = filepath, code = code)
+end
+
+provider_mux          = prefetch_provider_file("mux.jl")
```
travigd: Nit: can you format these without the spacing before the `=`s?

NHDaly: done, thanks!

NHDaly:
Ah, thanks for explaining. Yes, that is definitely a problem as well! 😁 

The ApplicationBuilder.jl package has _some_ infrastructure set up to let you list "assets" like these, which get copied into the final .app bundle, and then you have to manually add some code in your project that says basically `if _is_static_compiled_() @eval WebIO const packagepath = "assets/webio/packages"` or wherever you had them moved to. It's super hacky and dumb, but this is how it is for now.

However, that _doesn't_ work for things that are referenced in the `__init__()` function of a module, because the `__init__()` is run before your code gets a chance to override the path! So basically, the rule is that references to absolute paths from within `__init__()` functions are hostile to relocatability of Applications.

---------

That said, the ultimate best approach would probably be to make any assets like that into Artifacts through Julia's new Artifacts system, since those are totally path-agnostic and relocation-friendly, but I dunno if that would be overkill for something like what you're describing...

travigd:
I have no opposition to making assets into artifacts (though ideally, I'd like to keep compatibility with Julia 1.0 - I'm not sure if that's an issue?).

I don't really want to do that work myself though. :^)

NHDaly:
Yeah agreed. And i'd say there isn't _really_ much reason to. For those who care about it, the hacks i described above are mostly good enough for now...

shashi:
Nice PR! :-)

About frontend assets, @ranjanan once got an app to compile. The main idea is to keep the assets in some path and setup a const like @NHDaly showed, then AssetRegistry.register them in `__init__`, or just use them in Scopes so it's automatic.

PR #379, opened 2020-01-06
webio_serve(::Mux.App)

cstjean:
The docstring says:

```julia
  webio_serve(app, port=8000)

  Serve a Mux app which might return a WebIO node.
```

but, frustratingly, it does not currently accept a `Mux.app`. This PR trivially fixes this issue.


shashi:
Thanks!

PR #380, opened 2020-01-07
Remove `@info` output

nilshg:
As discussed with @travigd this message keeps popping up randomly and seems to make Juno overall more laggy (in particular in combination with `TableView`) so this is the nuclear option - NB I don't really know what the function does overall so this might well be a bad idea!

### DIFF
--- a/src/connection.jl
+++ b/src/connection.jl
@@ -114,7 +114,6 @@ function send_message(
         if isopen(connection)
             send(connection, msg)
         else
-            @info "Connection is not open." connection
             delete!(pool.connections, connection)
         end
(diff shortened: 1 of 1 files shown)


travigd:
> seems to make Juno overall more laggy

I don't know what that would be about. But it's fine to delete this info message.

nilshg:
The message seems to be emitted in batches of 5-6 at a time, and it feels to me like this happens when I do `showtable(df)` from TableView or some plots and that it introduces some lag before the result of my command is displayed. This is just my personal impression though I haven't in any way verified or timed this.

PR #385, opened 2020-01-24
Typo

Balinus:


travigd:
Test failure seems to be a weird transient TravisCI issue and only occurred on the Mac/Julia1.3 test.

travigd:
Thank you for catching this!

PR #392, opened 2020-02-25
allow Observables 0.3

piever:


travigd:
Can we automate this dependency update process?

PR #394, opened 2020-03-09
Show exception with stack trace on send_message error

yha:
Helps with JuliaGizmos/Interact.jl#290.
It doesn't solve that issue yet, since errors are still swallowed if produced in the initial position of the `@manipulate` slider.

### DIFF
--- a/src/connection.jl
+++ b/src/connection.jl
@@ -117,12 +117,10 @@ function send_message(
             delete!(pool.connections, connection)
         end
     catch ex
-        showerror(stderr, ex, stacktrace(catch_backtrace()))
-        println(stderr)
+        trace = sprint(showerror, ex, stacktrace(catch_backtrace()))
         @error(
             "An exception occurred while trying to send a WebIO message to a "
-                * "frontend:",
-            exception=ex,
+                * "frontend:\n$trace"
         )
         delete!(pool.connections, connection)
     finally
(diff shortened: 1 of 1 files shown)


travigd:
Thanks for the PR.

Can we instead just include this in the `@error` invocation?

```julia
trace = sprint(io -> showerror(io, ex, stacktrace(catch_backtrace())))
@error(
    ...,
    trace,
)
```

yha:
I tried that initially, but passing the trace as a second argument like this shows it as an escaped string with `\n`s and no newlines, which is pretty hard to read.
The new commit adds the trace into the first argument, which seems to work fine.

pfitzseb:
The "right way" to do this is to pass an `(exception, backtrace)` tuple to `@error`:
```
julia> try
       0//0
       catch e
       @error("0//0", exception=(e, catch_backtrace()))
       end
┌ Error: 0//0
│   exception =
│    ArgumentError: invalid rational: zero(Int64)//zero(Int64)
│    Stacktrace:
│     [1] __throw_rational_argerror(::Type) at ./rational.jl:19
│     [2] Rational at ./rational.jl:14 [inlined]
│     [3] Rational at ./rational.jl:21 [inlined]
│     [4] //(::Int64, ::Int64) at ./rational.jl:44
│     [5] top-level scope at none:2
```

yha:
Right, I somehow missed that going through the `@error` docs. Updated the PR.

travigd:
LGTM.

TravisCI failures are seemingly unrelated (though should be fixed).

Issue #416, opened 2020-05-29
directory serving not possible


hhaensel:
Currently, WebIO does not allow for serving of directories.

When I try `WebIO.deps2url(<somedirecory>)` I receive an error that the argument needs to be a valid file or url.

I propose to slightly modify `asset.jl` to allow for the serving of directories. This is a prerequisite to adapt PlotlyJS to include mathjax in its output.

I plan to file PRs for both modifications.

PR #417, opened 2020-05-29
add possibility to register directories

hhaensel:
This is the corrseponding PR to issue #416 

Currently, WebIO does not allow for serving of directories.

I propose to slightly modify asset.jl to allow for the serving of directories. This is a prerequisite to adapt PlotlyJS to include mathjax in its output.

To me it looks as if the design is already made for this purpose, but is not fully completed yet. I think one return was forgotten and I chose to use `SubString` to remove the initial path but include the separator.

### DIFF
--- a/src/asset.jl
+++ b/src/asset.jl
@@ -181,19 +181,21 @@ function path2url(path::AbstractString)
             :webio_pkg_urls,
         )
         return path
-    elseif startswith(path, "/assetserver/") && haskey(AssetRegistry.registry, path)
+    # check if file or containing directory has already been registered
+    elseif startswith(path, "/assetserver/") && haskey(AssetRegistry.registry, join(split(path, "/")[1:3],"/"))
         return path
-    elseif isfile(abspath(path))
+    elseif isfile(abspath(path)) || isdir(abspath(path))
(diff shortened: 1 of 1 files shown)


hhaensel:
I've added another commit, which does not cut the query string at a potential '?'.
It's even a bit shorter ;-)
But I wonder whether `baseurl[] * ` should not be omitted here as it already occurs in `AssetRegistry.getkey`. Instead it should probably be added in line 185 before `"/assetserver/"`. Please correct me, if I'm wrong.
One last question. Currently `AssetRegistry.isregistered()` computes the key everytime. Wouldn't it be better to check for the value instead, i.e.
```
isregistered(path) = abspath(path) in values(registry)
```
instead of
```
isregistered(path) = haskey(registry, getkey(path))
```
Getkey could be made faster along the same lines by first checking existing entries.

```
generatekey(path) =  baseurl[] * "/assetserver/" * bytes2hex(sha1(abspath(path))) * "-" * basename(path)

function getkey(path)
    key = findfirst(x -> x == abspath(path), registry)
    key === nothing ? generatekey(path) : key
end
```
I could file a PR for AssetRegistry as well, if this is welcome.
The main time killer, however, is `abspath()`, at least on my windows machine ...

PR #453, opened 2021-04-26
Fix misleading onjs() examples

green-nsk:
Looks like the expression in docs is not working as expected.


travigd:
This syntax is actually the preferred way to hook into observables in JavaScript. If you're having issues with it, please feel free to open an issue.

PR #457, opened 2021-06-09
JupyterLab 3 compatibility

travigd:
Closes #442.

PR #461, opened 2021-07-02
Update ijulia.md

qSeven:
I suggest adding an extra hint for adding WebIO for jupyter notebooks in addition to the one jupyter lab since this could save some debugging time.


travigd:
Closing in favor of #457 (still a WIP, but has obsoleted the `WebIO.install_jupyter_nbextension()` function)

PR #477, opened 2022-01-20
do not try to load external packages in Requires.jl required files

KristofferC:
This fixes the issue reported in https://github.com/JuliaLang/PackageCompiler.jl/issues/650


travigd:
Ty!

KristofferC:
Would be nice with a new release as well.

travigd:
Yes, I think this PR actually conflicts with the other requires.jl PR, so we might need to fix that up, but I'll get a release out very soon.

PR #478, opened 2022-01-20
rely less on Requires.jl

### LINKED ISSUE #476: Rethink @requires dependencies
The startup time for Blink.jl is almost entirely consists of the requires blocks here.

The WebSockets block is unnecessary because its a dependency anyway. Mux.jl is a small package that shares many dependencies. It's much faster to depend on it here than to include it in quires.

Blink.jl already depends on WebIO.jl - the code should just live in Blink.jl where it can be precompiled.

All up this reduces the package load time of Blink.jl (on my laptop) from 5 seconds to 1 second.

See: https://github.com/JuliaGizmos/Blink.jl/issues/288

If people agree I can PR Blink and WebIO with the changes.

rafaqz:
This PR reduces the use of Requires.jl.

- it moves WebSockets code out of Requires, its a direct dependency anyway
- Adds a dependency to Mux.jl. Its small and shares most deps, and really makes things faster having it precompiled.
- Removes Blink.jl completely. Blink depends on WebIO and can define its own WebIO methods, which will also be precompiled.

This needs a companion PR to Blink.jl, to do once this is approved. Overall this PR improves Blink.jl load time from 5s to 1s, on my laptop.

Closes #476

See https://github.com/JuliaGizmos/Blink.jl/issues/288


travigd:
I'll take a look at this this weekend.

travigd:
This looks decent to me. So no actual code changes are needed for Blink.jl? If that's the case, this can just be a patch release.

rafaqz:
Maybe you forgot about this one

https://github.com/JunoLab/Blink.jl/pull/201

rafaqz:
I cant get the tests running locally, the `npm` call doesn't spawn. But if you can approve the CI here we can see if it runs

rafaqz:
I dont really understand that documentation failure. Any ideas?

rafaqz:
Ok I think I've fixed this, if you can try running it again.

rafaqz:
bump

rafaqz:
Can you switch the github actions permissions over to "Require approval for first-time contributors"?

It's hard to debug the docs problems currently

rafaqz:
@travigd lets get this merged before I forget about it. There is a problem with building the docs on github actions. 

If you change the github actions permissions over as mentioned above I can get it passing. Approving every CI run on every push to JuliaGizmos PRs doesn't seem to be a good use of your time to me!

travigd:
Sorry for being incommunicado recently. My life has been hectic and I use Julia very little in my day-to-day work life nowadays.

The docs build is failing because the `build` step fails (or just isn't run?) in CI. There are a few possible solutions:
* Build it in the CI for docs. This would require install nodejs so it can run all the JS toolchain stuff.
* Change the `error(...)` in `__init__` to be a log warning (e.g., `@error`).
* Have the `bundle_key` ref be initialized lazily as a function:
  ```julia
  _bundle_key_ref = Ref{String}("")
  function _bundle_key()
    current = _bundle_key_ref[]
    if current == ""
      # try to initialize it here
    end
    return current
  end
  
  # later
  string(base, _bundle_key())
  ```

I'm partial to the third option.

PR #484, opened 2022-02-22
Use Artifacts

pankgeorg:


deps/bundlepaths.jl
```diff
@@ -50,8 +51,14 @@ function download_js_bundles()
...
 
     mkpath(BUNDLES_PATH)
-    download_bundle("core", CORE_BUNDLE_PATH, CORE_BUNDLE_URL)
-    download_bundle("generic-http", GENERIC_HTTP_BUNDLE_PATH, GENERIC_HTTP_BUNDLE_URL)
-    download_bundle("mux", MUX_BUNDLE_PATH, MUX_BUNDLE_URL)
-    download_bundle("blink", BLINK_BUNDLE_PATH, BLINK_BUNDLE_URL)
+    bundle_artifact_path = artifact"web"
+    for asset in readdir(bundle_artifact_path)
+        @debug("Copying", bundle_artifact_path, asset, BUNDLES_PATH)
+        cp(joinpath(bundle_artifact_path, asset), joinpath(BUNDLES_PATH, asset))
```
pfitzseb, line 57:
```suggestion
        cp(joinpath(bundle_artifact_path, asset), joinpath(BUNDLES_PATH, asset); force = true)
```

Artifacts.toml
```diff
@@ -0,0 +1,6 @@
+[web]
+git-tree-sha1 = "050430be38d442e70243cb830455185b9fd0f5bd"
+
+    [[web.download]]
+    sha256 = "00debe8457bc840c81e204c482bada9098446aa12f5009c82b5c3b8b0281cebd"
+    url = "https://github.com/pankgeorg/WebIO.jl/files/8111257/bundle.tar.gz"
```
pfitzseb, line 6: Hm, would be good to get this automatically generated and hosted in this repo, I think.

pankgeorg: The idea here is to make a release, add this there, do an extra commit with the link from the release assets and register that. Think of this as bootstrapping; are you thinking of something neater?

pfitzseb: I was thinking of doing that automatically, but I guess it's not worth the effort, considering the (hopefully) low number of future releases.

PR #497, opened 2022-10-31
Update use of Observables listeners to current interface

halleysfifthinc:
Non-MWE:
- In IJulia, with PlotlyJS, an observer on a plot event (e.g.
  "hover", "click", etc) gave the following error when updating/being run:

```julia
KERNEL EXCEPTION
MethodError: objects of type Pair{Int64, Any} are not callable

Stacktrace:
  [1] #invokelatest#2
    @ ./essentials.jl:729 [inlined]
  [2] invokelatest
    @ ./essentials.jl:726 [inlined]
  [3] set_nosync(ob::Observable{Dict{Any, Any}}, val::Dict{String, Any})
    @ WebIO ~/.julia/dev/WebIO/src/scope.jl:355
  [4] dispatch(ctx::Scope, key::String, data::Dict{String, Any})
    @ WebIO ~/.julia/dev/WebIO/src/scope.jl:368
  [5] dispatch_command(conn::WebIO.IJuliaConnection, data::Dict{String, Any})
    @ WebIO ~/.julia/dev/WebIO/src/messaging.jl:104
  [6] dispatch(conn::WebIO.IJuliaConnection, data::Dict{String, Any})
    @ WebIO ~/.julia/dev/WebIO/src/messaging.jl:81
  [7] (::WebIO.var"#97#98"{WebIO.IJuliaConnection})(msg::IJulia.Msg)
    @ WebIO ~/.julia/dev/WebIO/src/providers/ijulia.jl:21
  [8] comm_msg(sock::ZMQ.Socket, msg::IJulia.Msg)
    @ IJulia.CommManager ~/.julia/packages/IJulia/AQu2H/src/comm_manager.jl:134
  [9] #invokelatest#2
    @ ./essentials.jl:729 [inlined]
 [10] invokelatest
    @ ./essentials.jl:726 [inlined]
 [11] eventloop(socket::ZMQ.Socket)
    @ IJulia ~/.julia/packages/IJulia/AQu2H/src/eventloop.jl:8
 [12] (::IJulia.var"#15#18")()
    @ IJulia ./task.jl:484
```

I narrowed the problem down to an incorrect assumption about what is
returned from `Observables.listeners`. The previous expectation was an
array of functions, the reality is a `Vector{Pair,Function}`.

WebIO#master doesn't pass tests on my machine, so I can't say with certainty that this doesn't cause any problems, but this does fix the example issue above.

I removed the conditional test for a ~2yo Observables version and removed the corresponding compat entry.


MasonProtter:
@halleysfifthinc I just ran across the same thing! I think there's two further changes that you need to make here, so I've opened a PR here: https://github.com/halleysfifthinc/WebIO.jl/pull/1

PR #500, opened 2022-11-02
Switch from WebSockets.jl to HTTP.jl

fonsp:
HTTP.jl recently had a big overhaul of the WebSockets code, and it is now very stable, and tested against the [autobahn test suite](https://github.com/crossbario/autobahn-testsuite). HTTP.jl also released 1.0, which means that we want the whole Julia ecosystem to make the upgrade.

WebIO.jl currently uses WebSockets.jl, which depends on an old version of HTTP.jl. They recently [changed the compat on 	`master`](https://github.com/JuliaWeb/WebSockets.jl/commit/7688eea76361cf1ccfe8481c413958d1e99bfce4) for HTTP to include `1`, but without making compatibility changes, and the package now [fails to import](https://github.com/JuliaWeb/WebSockets.jl/issues/179). In my view, the best way for dependents of WebSockets.jl to stay up-to-date is to switch to HTTP.jl, which is more stable, performant, actively maintained, etc.

Because of #499, I was not able to run or test this code locally, so I need someone else to finish this PR. 💕

PR #502, opened 2022-11-09
Set minimum Julia version to 1.6

fonsp:
Since #497, we only support Observables.jl 0.5 and up, which only supports Julia 1.6 and up. I updated the GitHub Actions to only run tests on Julia 1.6 and up, and updated the Project.toml to drop versions that we can't test.


fonsp:
Future compatibility with HTTP.jl 1.0 (e.g. #500) will also require Julia 1.6 and up.

PR #503, opened 2022-12-10
WebSockets 1.6.0

anj00:
WebSockets support 1.6.0
As interface to websockets didn't change, there is no code change. 

I also run this version of WebIO against not yet released Blink (https://github.com/JuliaGizmos/Blink.jl/pull/295) and Blink tests pass. 

yes, in the PR the Blink is not actually using WebSockets, but unless WebIO moves to latest WebSockets, Blink cannot upgrade to latest HTTP


pfitzseb:
Thanks!

PR #520, opened 2026-03-22
update node version to fix npm dependency installation

hhaensel:


hhaensel:
The remaining errors are fixed in #519

hhaensel:
@shashi would you consider merging this PR first and then #519?

hhaensel:
I cherry-picked a commit from #519.  Strangely, I see julia 1.12 failing while 1.6 passes. In my local wsl ubuntu environment the tests pass

